PR #1294, opened 2015-01-11
Added two python books

quake004:
I suggest writing down what books are open source so people seeking to translate a book can find one quickly.

### DIFF
--- a/free-programming-books-es.md
+++ b/free-programming-books-es.md
@@ -199,6 +199,8 @@
 
 
 ### Python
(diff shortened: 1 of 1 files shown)


vhf:
I think it's not worth the effort.

eshellman:
I have also thought about doing this; I would help out if we could get agreement on how to go about it. Perhaps the best way to approach it is to create an issue.

diggydoc4u:
I.think creating a issue is the best way to really test this.

borgified:
going to have to put it in alphabetical order later. thx for your contribution!

PR #1297, opened 2015-01-16
Adding CC-By-SA requirements for attribution

gortok:
Adding the attribution requirements as stated in this blog post: http://blog.stackoverflow.com/2009/06/attribution-required/.  Only added it to the actual markdown file that was created from the original answer on Stack Overflow.

 A reading of the attribution requirements does not seem to indicate that a sole reference in the readme.md is enough to satisfy the CC-By-SA requirements for attribution.


vhf:
Many thanks.

PR #1337, opened 2015-02-08
Add a link to Modern C book (as pdf).

jbranchaud:


vhf:
I'm not sure this book is free. Could you please double-check?

eshellman:
It's not free licensed, but it seems clear the author is ok with links to it:

> This is a preliminary version of this book compiled on February 10, 2015.
> It contains feature complete versions of Levels 0, 1 and 2, and the table of contents
> already gives you a glimpse on what should follow in the two missing ones.
> You might find a more up to date version at
> http://icube-icps.unistra.fr/index.php/File:ModernC.pdf (inline)
> http://icube-icps.unistra.fr/img_auth.php/d/db/ModernC.pdf (download)
> You may well share this by pointing others to my home page or one of the links above.
> Since I don’t know yet how all of this will be published at the end, please don’t distribute the file itself, yet:
> All rights reserved, Jens Gustedt, 2015

vhf:
Thanks

PR #1339, opened 2015-02-09
Adding books to the JS section

willricketts:


vhf:
The first link is already present in the free-programming-books file, the second in the interactive file.

PR #1345, opened 2015-02-11
Remove Railscasts from books.

tejasbubane:
Remove Railscasts from list of free books since it is already in the screencasts list in `free-podcasts-screencasts-en.md`.


vhf:
Thanks!

PR #1397, opened 2015-02-26
New Book added - Problem Solving with Algorithms and Data Structure

ashwani27493:
Problem Solving with Algorithms and Data Structure - Brad Miller and David Ranum, Luther College


neuhaus:
the full book title is "Problem Solving with Algorithms and Data Structures Using Python"

ashwani27493:
Corrected.

PR #1425, opened 2015-03-26
Remove deadlink for Java book.

ajkamel:


vhf:
Does not look dead to me.

ajkamel:
If you click through to the next set of url's contained within the index they all sql fetching error's.  I tried finding an alternate source however I was unable to do so.

PR #1429, opened 2015-04-03
Added indexing to the JavaScript Frameworks list.

chrisspruck:
Anchor links having . in them like #angular.js don't seem to work correctly, so I looked at other indices in this repo and patterned this after those, where the anchor name had JS capitalized, i.e. AngularJS instead of Angular.js. I'm happy to redo this in a different pattern if desired. The anchors all work as written now, and I tested in http://dillinger.io/ and on the live page in my branch. Thanks!

### DIFF
--- a/javascript-frameworks-resources.md
+++ b/javascript-frameworks-resources.md
@@ -1,4 +1,15 @@
-## Angular.js
+### Index
+* [Angular.js](#angularjs)
+* [Backbone.js](#backbonejs)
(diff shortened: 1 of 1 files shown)


chrisspruck:
Hello - I'm curious if there was any reason this wasn't accepted without comment. I followed other files that had an index and figured this would be useful as the file length got longer. Should I have asked if this was ok to do? Thanks!

PR #1440, opened 2015-04-13
修改链接地址

0532:


weakish:
Hmm, this updates the link. (The original link will redirect though.)

PR #1457, opened 2015-05-04
Added Free Courses

Akhilan:
Adding a file with Free Video Courses and Channels of Popular Universities


borgified:
would these videos fit here instead? ```free-podcasts-screencasts-en.md```

onebree:
I think "playlists" is the wrong word, since you are merely linking to the channel's main page. I believe free videos is the correct file. Instead, I would create a section "Universities on YouTube", or something...

PR #1491, opened 2015-06-08
Added some podcasts

oraac:
added Programming Throwdown & Software Engineering Radio


MJ-Vakili:
Is the address correct?

http://http://www.programmingthrowdown.com/
http://http://www.se-radio.net/

MJ-Vakili:
@vhf  Is the address correct?
http://http://www.se-radio.net/
http://http://www.programmingthrowdown.com/

vhf:
Nope you're right. I'll fix it asap. Thanks for the catch.

PR #1516, opened 2015-06-28
Added a fortran book.

saaketp:


vhf:
I don't think this book is free. It says Springer, and Springer usually does not give their publications away.

saaketp:
OK, sorry I didn't check.
I removed that entry and added another book which is free.

PR #1527, opened 2015-07-02
added "How to Design Programs, Second Edition"

seralf:
from: http://htdp.org/


vhf:
Since it's a book, could you please add it in the free-programming-books.md instead?

Thanks a lot.

seralf:
sorry I added there by mistake, and it is already cited in the right file! :-)

PR #1564, opened 2015-08-09
Add business free  courses

iKenshu:


onebree:
I think this should be either under MISC, with its own set of bullets, OR moved to a different awesome list. I prefer the latter, because this list is for _programming_, and not business or hiring manuals.

:-1:

iKenshu:
Yea, that was I thought but i don't know, had to try. I think could be really useful for some people

vhf:
Thanks for the PR and comments. I'd like to keep this list programming-oriented. :)

PR #1580, opened 2015-08-29

wuzhouhui:
add The AWK Programming Language, Chinese Edition.
The translation is ongoing, but I will hold on.

add The Practice of Programming, by Brian W. Kernighan and Ron Pike.

### DIFF
--- a/free-programming-books-zh.md
+++ b/free-programming-books-zh.md
@@ -7,6 +7,7 @@
 	* [数据库](#数据库)
     * [智能系统](#智能系统)
     * [正则表达式](#正则表达式)
+* [AWK](#awk)
 * [C](#c)
 * [C++](#c-1)
 * [CoffeeScript](#coffeescript)
@@ -86,6 +87,8 @@
 * [正则表达式30分钟入门教程](http://deerchao.net/tutorials/regex/regex.htm)
 
 
+### AWK
+* [AWK 程序设计语言](https://github.com/wuzhouhui/awk)
 
 ### C
 * [C 语言常见问题集](http://c-faq-chn.sourceforge.net/ccfaq/ccfaq.html)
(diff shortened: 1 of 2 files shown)


free-programming-books.md
```diff
@@ -560,6 +560,7 @@ Original Source: [Free Programming books](http://stackoverflow.com/revisions/392
 #### Programming Paradigms
 * [Introduction to Functional Programming](http://www.cl.cam.ac.uk/teaching/Lectures/funprog-jrh-1996/) (class lectures and slides)
 * [Type Theory and Functional Programming](https://www.cs.kent.ac.uk/people/staff/sjt/TTFP/)
+* [The Practice of Programming](https://github.com/wuzhouhui/tpop) (PDF) *(by Brian W. Kernighan and Ron Pike)*
```
borgified: https://en.wikipedia.org/wiki/The_Practice_of_Programming says that the book has already been translated into 12 languages maybe one of them is Chinese?

Also, I dont know if it is ok to make a translation of a [copyrighted](http://www.informit.com/store/practice-of-programming-9780201615869) book and slap on an [MIT license](https://github.com/wuzhouhui/tpop/blob/1717cfd3da86bf1268106ace9847386c3e41459f/LICENSE) on it is an OK thing to do. 

Need to consult @vhf

free-programming-books-zh.md
```diff
@@ -86,6 +87,8 @@
 * [正则表达式30分钟入门教程](http://deerchao.net/tutorials/regex/regex.htm)
 
 
+### AWK
+* [AWK 程序设计语言](https://github.com/wuzhouhui/awk)
```
borgified: https://en.wikipedia.org/wiki/The_AWK_Programming_Language isnt in the public domain, yes you are translating it but i have to defer to @vhf to decide whether we can include.

free-programming-books.md
wuzhouhui: The book I added is English edition, not Chinese. Every word on this book is typped by me, and I
also don't know the matters about copyright.

onebree:
I would have to give a :-1: for this. Copyrights for published works are very tricky to deal with. It isn't like Creative Commons. Copyrights usually prohibit any distribution or altering of the work without explicit permission. Personal use is fine, since you own that copy of the book. If possible, I would submit a bug report to the errata section of the book's main page. (If not, find a contact email.) Maybe they could make changes for a future edition.

While you "slapped on" an MIT license, your work is technically a derivative of the original, and therefore you must abide by _the original's license_. I would suggest moving your repo to a private one on Bitbucket (free), before Github attempts to take it down. Wouldn't want any harm done on your account.

PR #1597, opened 2015-09-07
Correct minor typos

palmerev:
change "A" to "The" in "The Haskell School of Music" to match the title on the site.
change Django version number on "Test-Driven Web Development with Python" to 1.8. As the author notes [here](http://chimera.labs.oreilly.com/books/1234000000754/pr02.html#ftn.id619210), he updated the book for 1.8 in Spring 2015, since it's an LTS version.


borgified:
:+1:

PR #1603, opened 2015-09-11
D3 Tips and Tricks - updated formats and author

tomexx:
I want to find out how to do pull requests. It is just a small update.


onebree:
I do not think it is necessary to include all those formats. I would suggest to replace that with `(leanpub)` or something.

tomexx:
Updated

PR #1604, opened 2015-09-11
PHP book: Using Libsodium in PHP Projects

peterkokot:
This patch adds free, open source, online book for PHP: Using Libsodium in PHP Projects - a guide to using libsodium PHP extension for modern, secure, and fast cryptography.


borgified:
looks good thank you

PR #1612, opened 2015-09-21
Add Learn Enough Command Line to Be Dangerous

mistermatze:


onebree:
:+1:, although I would note that the book is still currently a draft. I am sure the final edition will still be free. Noting that it is a draft tells someone that all information is not final and may contain errors.

PR #1614, opened 2015-09-23
Replace dead SO link with archived one

onebree:
Found this dead link still on the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,6 @@
 [View the list](https://github.com/vhf/free-programming-books/blob/master/free-programming-books.md)
 
 ## Intro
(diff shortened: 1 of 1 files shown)


onebree:
@vhf The reSRC project is no longer active, but still present on the README. I wanted to link to the repo, but could not find any material that would be helpful, besides tests and assets. What should we do in this case?

vhf:
Regarding reSRC, feel free to remove any reference to it from the README and add these commits to this PR, I'll merge it. :)

Thanks a lot!

onebree:
@vhf made the second commit :-)

Besides web archive, can we still access material from resrc.io? I couldn't find anything upon a quick browse of the repo

vhf:
Unfortunately not...

PR #1622, opened 2015-10-02

dreamtocode:
Add book of Raspberry Pi


vhf:
I don't think this is a free book. Could you provide a copyright/left notice showing that this book is indeed free? Thanks!

dreamtocode:
@vhf Oh no I don't think that there is any..

PR #1624, opened 2015-10-02
Alphabetizing the german book index

eTomate:
Sorted the books by alphabet


vhf:
Great work, thanks a lot!

PR #1627, opened 2015-10-02
Add RailsGuides (Railsガイド) and correct Rails Tutorial in Japanese

yasulab:
I'm co-founder of both projects and would like to add/correct the info :)


yasulab:
Thanks for merging this ;)

PR #1629, opened 2015-10-04
Added C++ Core Guidelines

asm0x:
Citing authors: "The C++ Core Guidelines are a set of tried-and-true guidelines, rules, and best practices about coding in C++"


vhf:
Thanks! Is this really a book though?

asm0x:
In my opinion. In docs directory is a few pdf files. These docs are still updated, so I gave link to github repository.

horar:
That pdf files are mostly from CppCon; anyway the document/page refered here is worth to mention. Yes, it's still evolving but it is (should be) for many developers as guido how to write the C++

PR #1636, opened 2015-10-14
Add Postgres official documentation

jwworth:
This is _the_ resource for anyone using Postgres.

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -1502,6 +1502,7 @@ See also [TeX](#tex)
 
 
 ### PostgreSQL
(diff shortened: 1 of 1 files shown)


vhf:
I'm hesitating to include this link because it's more a documentation than a book.

jwworth:
Fair point. It's organized like a book, and I saw other official docs in the list already (Python, Django), so I thought it might be useful. Your call!

vhf:
Hmm, I see that now. I think I would prefer to also remove the other official docs. I'll leave this issue open as a reminder, and also because I have not made up my mind yet.

PR #1643, opened 2015-10-17
Added Java for Complete Beginners in courses

Tohoma:
I believe the course is compatible with the criteria


vhf:
Indeed, thanks!

PR #1665, opened 2015-10-27
Added AngularJs and javascript category adjusted in pt-br

gabrielgomesferraz:
I added the category angularjs and think better left javascript only what's pure javascript and when it involves node, jquery and other api divide the category is thus better understanding, even if all are javascript.

### DIFF
--- a/free-programming-books-pt_BR.md
+++ b/free-programming-books-pt_BR.md
@@ -14,9 +14,9 @@
 * [Java](#java)
 * [JavaScript](#javascript)
 * [jQuery](#jquery)
-* [Node.js](#node.js)
 * [LaTeX](#latex)
 * [Lua](#lua)
+* [Node.js](#node.js)
 * [PHP](#php)
 * [Python](#python)
 * [RegEx](#regex)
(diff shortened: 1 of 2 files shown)


onebree:
Your sources needed to be alphabetized within each category.

vhf:
Thank you for this proposition.

I don't really like the idea of mixing frameworks/libraries with languages. It's not a matter of putting them in the same file, rather a matter of not putting them at the same hierarchical level in the index/in the file.

In `free-programming-books.md` we used two approaches. For some languages (e.g. Java), we filed the frameworks/libraries as sub-level titles, both in the content and in the index. For JS, we created a separate file.

What would think of the idea of creating a file named `javascript-frameworks-resources-pt_BR.md` on the same model as `javascript-frameworks-resources.md`, and make a link in the JavaScript category of `free-programming-books-pt_BR.md` pointing to that file?

I think this would be the most consistent way of doing things. I would be happy to discuss this idea here with anyone interested!

gabrielgomesferraz:
Great idea, I would be willing to do it quietly.

gabrielgomesferraz:
Ready, the changes were carried out, please check if everything is okay.

vhf:
Thanks!

PR #1666, opened 2015-10-27
Add Red Gate .NET resources

jrg1381:
We have a free community-sourced book of .NET performance tips ("52 secrets...") and two older tips-and-tricks compilations.


jrg1381:
As you can see, I merged master into my PR.

PR #1667, opened 2015-10-27
Add ebook The Little Go Book

jeanmatheussouto:
Added ebook `The Little Go Book`


onebree:
IMO, you should replace the PDF link to that of the official repo, which will be constantly updated.

https://github.com/karlseguin/the-little-go-book

jeanmatheussouto:
Really better @onebree :+1:

PR #1669, opened 2015-10-27

rvu95:
added unix power tools


onebree:
I am not sure how I feel about this link. The site denotes credit and copyright to OR, but I found the notice below. 

> This CD-ROM is intended for use by one individual. As such, you may make copies for your own personal use. However, you may not provide copies to others, or make this reference library available to others over a LAN or other network. You may not reprint, offer for sale, or otherwise re-use material from this product without the explicit written permission of O'Reilly & Associates, Inc. If you would like to discuss licensing for multiple users, contact service@oreilly.com.

vhf:
Agreed; does not look like a free book.

PR #1696, opened 2015-11-09
Adding a section for Angular 2 and a link to a good resource

rafaelaudy:
Adding a link to the Angular 2 resource compilation repository.
Probably the best link for people learning this new framework. :)


onebree:
Is Angular 2 related to Angular JS? Is it just a version upgrade? If so, I think it should go to the already present AngularJS section.

If not, then never mind this comment. :-)

rafaelaudy:
They are quite different but I guess you are right.
I am going to open another pull request with that link under Angular.js :)

PR #1698, opened 2015-11-10
Add A Brief Introduction to Neural Networks

natehefner:


vhf:
Please replace this URL with the one suggested by the author: http://www.dkriesel.com/en/science/neural_networks

This way, the link won't break when the author updates the PDF URL.

natehefner:
Ah, my fault for overlooking that. Fixed!

vhf:
Thank you! One last thing: please remove the PDF mention, now that the link does not point to a PDF file.

natehefner:
Done, thanks!

PR #1700, opened 2015-11-11
Fix and add Japanese scala contents

yutaono:


vhf:
Thanks! Could you please put "Effective Scala" in alphabetical order?

yutaono:
Sure. I fixed it :)

PR #1701, opened 2015-11-12
Adopt a Code of Conduct

vhf:
I think it's a shame that the second most popular github repository does not have a code of conduct and I intend to fix this. We never needed a code of conduct, we don't want to need a code of conduct and that's why we need a code of conduct.

I submit this as a PR instead of committing directly to master because I want to allow some debate on this issue. I will merge this PR in a week or so, if there is no good case against it.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,50 @@
+# Contributor Code of Conduct
+
+As contributors and maintainers of this project, and in the interest of  
+fostering an open and welcoming community, we pledge to respect all people who  
+contribute through reporting issues, posting feature requests, updating  
+documentation, submitting pull requests or patches, and other activities.
+
+We are committed to making participation in this project a harassment-free  
+experience for everyone, regardless of level of experience, gender, gender  
+identity and expression, sexual orientation, disability, personal appearance,  
+body size, race, ethnicity, age, religion, or nationality.
+
+Examples of unacceptable behavior by participants include:
+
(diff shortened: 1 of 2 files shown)


onebree:
Although this repo is just a set of links, I think a COC is a great idea. Like you said, this is a popular repo, that is often hit by spammers. While a COC will not prevent spammers, at least rules are still implemented.

1. Why did you choose to adopt this particular COC?
2. The rendered Markdown treats each LOC as a new line of text. I think the 2 spaces following each line causes this. If the COC becomes a markdown file, I think the text should flow as expected on a webpage.

vhf:
Thanks for your input.

1. I had a look at the most popular COCs and chose this one. [This resource](http://geekfeminism.wikia.com/wiki/Code_of_conduct_evaluations) helped me. This PR is also here to discuss if the proposed COC is the one we should adopt. I don't have strong feelings towards this particular COC and am open to dialogue.
2. Fixed - thanks!

onebree:
1. I do not have a strong opinion towards any one COC, either. Honestly, I did not take the time to read every line of this one, because, like mentioned before, we are just a repo of links. COC really comes into play during conversations, which are actually handled well already.
2. You did not need to write 1 line per paragraph. Markdown renders fine if a line (but mid-paragraph) ends with 0-1 spaces. Otherwise, it is treated as a line break. But hey, whatever :-) Still looks good.

eshellman:
Thanks for adding this.

PR #1705, opened 2015-11-17
Added a new Ruby repository

sarithats:


vhf:
Thanks. This is not what I would consider a book, so I suggest you add it to this list instead: https://github.com/markets/awesome-ruby

PR #1729, opened 2015-12-02
Add React Primer Draft by Mike Chau

jwworth:
This is a popular React primer and should be included in this list.


vhf:
Good, thanks!

PR #1730, opened 2015-12-03
Add ClojureScript Syntax in 15 Minutes.

jbranchaud:


vhf:
Thank you! I think we'll pass on this one. It's more a cheat sheet than a book.

Would probably be a wonderful addition to the awesome clojurescript list though!

PR #1733, opened 2015-12-04
Fix improperly capitalized "O"

ClimbsRocks:
The letter "o" in "COmbinatorics" was improperly capitalized. 

I checked the link, and they have "Combinatorics" without the "o" capitalized on their website, so this is just a typo. 

Here is a screenshot of the change:
screen shot 2015-12-04 at 12 24 44 pm


vhf:
Thanks!

PR #1736, opened 2015-12-09
Add an Om Tutorial in a new Om section for JavaScript frameworks.

jbranchaud:


vhf:
Thanks! Is it really a JavaScript framework resource?

jbranchaud:
I wasn't sure where to put it, but JS Frameworks seemed like a possibility since Om is a frontend (ClojureScript) framework that compiles to JavaScript. Thoughts on where would be a better place?

vhf:
Oh, if Om's sole purpose is to generate JS, why not! Thanks!

PR #1758, opened 2015-12-27
ko : add perl in index, include node.js in js.

openingnow:
* Add perl in index
* Include node.js in javascript section as in English document.


vhf:
Thanks!

PR #1764, opened 2015-12-29
ko : Add swift section with 1 document

openingnow:
ko : Add swift section with 1 document

It looks like I'm making PR only after 10 minutes after my previous PR accepted :)


vhf:
Great, thanks!

PR #1765, opened 2015-12-30
Added fpc ref guide; removed broken NV gems2 pdf

0xCD:


vhf:
Thanks! Please add ` (PDF)` after the link.

0xCD:
@vhf done.

PR #1791, opened 2016-01-12
Change links to comply with author request

eshellman:
Ian Parberry, the author, requests that links point to a license page
leading to his books rather than directly to pdf. See
https://github.com/vhf/free-programming-books/issues/1769#issuecomment-1
70046046


vhf:
Thanks!

PR #1792, opened 2016-01-12
Add free, beginner focused books from Launch School

knwang:
We've written a few beginner focused books that are openly available.


vhf:
Great, thanks!

PR #1806, opened 2016-01-21
add "Programming OpenResty" to section "OpenResty"

spacewander:
A book about an Nginx-based web development platform called OpenResty.


vhf:
Thanks!

PR #1810, opened 2016-01-22
Added two new titles to the Mathematics category.

impr0t:
Added:

Basic Algebra, Digital Second Edition (Anthony W. Knapp)
Advanced Algebra, Digital Second Edition (Anthony W. Knapp)


vhf:
Thanks!

http://www.math.stonybrook.edu/~aknapp/download.html

PR #1811, opened 2016-01-22
removed "Algorithms" (DP+V)

eshellman:
According to Prof. Papadimitriou, the pdf we were linking to (as well as others on the internet) are not authorized by the current copyright owner, McGraw Hill.


vhf:
Thanks!

PR #1813, opened 2016-01-23
Added "Best Ruby Practices" to list

jcoady9:
A github repo with information about best practices for the Ruby language.


vhf:
It's not really a book (yet?).

You should definitely add it to the awesome-ruby repo though!

PR #1819, opened 2016-01-24
replace unrelated party link with internet archive link

eshellman:
Judgement call. This is a 3Com "white paper". Though 3Com has copyright, white papers are typically freely distributed to advance a technology. It doesn't appear as if Prof. Ricci has anything to do with the white paper. The ooriginal link was likely lost on 3com's acquisition. I suggest replacing it with https://web.archive.org/web/20080401000000*/http://www.3com.com/other/pdfs/infra/corpinfo/en_US/501302.pdf
because if the owner wants to remove it they can do so.


vhf:
:+1:

PR #1827, opened 2016-01-31
Added advent of code. Closes #1773

### LINKED ISSUE #1773: Add Advent of Code to problem sets doc
http://adventofcode.com/

I would do this myself, but I am unsure which section it would go under.

While it was initially competitive, the public leader board is finished. I think "problem sets" is the better choice, especially since the challenges operate similarly to Project Euler.

What do others think?

kiriappeee:


onebree:
Why was this closed?

kiriappeee:
Sorry about that. Closed the issue because I added two commits that are unrelated to each other to a single pull request. Wanted to correct that. If this is ok though super happy to see it being merged in :smiley:

vhf:
Thanks!

PR #1866, opened 2016-03-07
Add Cryptography: An Introduction - Nigel Smart

triggerNZ:


vhf:
http://isup.me/www.cs.umd.edu : 

> It's not just you! http://www.cs.umd.edu looks down from here.

eshellman:
https://archive.org/download/CryptographyAnIntroductionNPSmart/CryptographyAnIntroductionNPSmart.pdf

PR #1868, opened 2016-03-08
Removing  dead links from Django

netofire:


netofire:
Removing dead links from Django

vhf:
Thanks!

PR #1878, opened 2016-03-14
Fixed ("C" to "C#")

sheeit:
Trailing hash signs (#) in headers are ignored. I fixed that with a HTML escape sequence.


vhf:
:+1:

PR #1889, opened 2016-03-26
Add Pro HTML5 Programming to HTML / CSS List

tidupls:


vhf:
Thanks! I think it would make more sense to link to <http://apress.jensimmons.com/v5/pro-html5-programming/ch0.html>. What do you think?

tidupls:
Totally agree. I changed the link.

PR #1899, opened 2016-04-02

Deathnerd:
Added Memory as a Programming Concept in C and C++ by Frantisek Franek

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -811,6 +811,7 @@
 * [Essential C](http://cslibrary.stanford.edu/101/EssentialC.pdf) (PDF)
 * [Learn C The Hard Way](http://c.learncodethehardway.org/book/)
 * [Modern C](http://icube-icps.unistra.fr/img_auth.php/d/db/ModernC.pdf) (PDF)
+* [Memory as a Programming Concept in C and C++](http://cs.ucsb.edu/~koc/cs32/docx/memorybook.pdf) - Frantisek Franek (PDF)
 * [Object-Oriented Programming With ANSI-C](http://www.planetpdf.com/codecuts/pdfs/ooc.pdf) (PDF)
 * [Programming in C](http://ee.hawaii.edu/~tep/EE160/Book/PDF/) - Bharat Kinariwala & Tep Dobry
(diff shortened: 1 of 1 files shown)


Deathnerd:
Original find for this goes to /u/tomizzo11 from the /r/learnprogramming subreddit ([Original post here](https://www.reddit.com/r/learnprogramming/comments/4d1etw/cc_program_memory_easily_the_most_helpful_book/)). 

I went ahead and made a pull request, but I have two issues with this and wanted to explain further here:

1. The link I provided is, to my knowledge, a re-hosting and not the original location of the work. The license on it (Cambridge University Press) is sparse: 

> This book is in copyright. Subject to statutory exception and to the provisions of relevant collective licensing agreements, no reproduction on any part may take place without the written permission of Cambridge University Press.

1. (continued) I feel as though submitting a link to a pre-existing copy is not infringing on this since it is not reproducing the work. I could be wrong though. I'm a programmer, not a lawyer lol.

2. It covers two languages (C and C++). My pull request has added the same link to both sections. This does not feel DRY to me, even though it is applicable to both languages. If this kind of duplication is to be avoided, perhaps it should only be in the C section since (almost) anything you can do in C can be carried over to C++ (I think. Don't quote me on that).

vhf:
Thanks for your contribution. I'll take a close look at the license
situation and decide, no worries! :)

Deathnerd:
Thanks for the speedy response. I did a little further Googling and it appears this is not a free book and is still being actively monitized [here](http://www.cambridge.org/us/academic/subjects/computer-science/programming-languages-and-applied-logic/memory-programming-concept-c-and-c) and [here](http://www.cambridge.org/us/academic/subjects/computer-science/programming-languages-and-applied-logic/memory-programming-concept-c-and-c). However, my original link is the second result on Google and it appears it's being shared many more places, though that doesn't mean we should do the same if it's still breaking copyright.

vhf:
If it's still sold on at least one legit website I'd prefer not to include
it. Thanks though!

Deathnerd:
I agree. I'll be closing this PR then.

PR #1910, opened 2016-04-08
Add "The Docker Book"

joelhandwell:
"The Docker Book is for developers, sysadmins, and DevOps-minded folks who want to implement [Docker™](https://www.docker.com/) and container-based virtualization." -- from Forword


tedyoung:
That book is *not* free (see the book's website at http://www.dockerbook.com/), so that's a pirated copy (which is also outdated).

Issue #1802, opened 2016-01-20
Dead links


ajcox:
None of these links are currently working.

Building and Testing with Gradle
Cadmus: A Primer in Java
J Reference Card
Learn Lisp The Hard Way
Learning Go
Object-Oriented Programming with Objective-C
PC-BSD® Users Handbook
Prolog Experiments in Discrete Mathematics, Logic, and Computability
Prolog Tutorial
Theory of Computation

onebree:
@ajcox Can you please add the hyperlinks to your comment above? While they may be dead, it will be easier to search for archives or reposts of.

ajcox:
Sure!

- Building and Testing with Gradle 
 - <https://gradle.org/servicesregistered-access?content=books%2Fbuilding-and-testing%2F>

- Cadmus: A Primer in Java 
 - <http://java.divshot.io/>

- J Reference Card 
 - <http://www.jsoftware.com/jwiki/HenryRich?action=AttachFile&do=view&target=J602_RefCard_color_letter_current.pdf>

- Learn Lisp The Hard Way 
 - <http://learnlispthehardway.org/book/>

- Learning Go 
 - <https://archive.miek.nl/projects/learninggo/index.html>

- Object-Oriented Programming with Objective-C 
 - <https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/OOP_ObjC/OOP_ObjC.pdf>

- PC-BSD® Users Handbook 
 - <http://wiki.pcbsd.org/index.php/PC-BSD%C2%AE_Users_Handbook>

- Prolog Experiments in Discrete Mathematics, Logic, and Computability  - <http://web.cecs.pdx.edu/~jhein/books/PrologLabBook09.pdf>

- Prolog Tutorial 
 - <http://www.cpp.edu/~jrfisher/www/prolog_tutorial/contents.html>

- Theory of Computation 
 - <http://www.cse.iitd.ernet.in/~ssen/csl705/admin705.html>

nickbarry:
The Express book also isn't available at the listed link:
Express.js Guide - Azat Mardanov (Leanpub)
https://leanpub.com/express/read

Hilariously, I see that the book is for sale on Amazon for an affordable $8,734.76:
![image](https://cloud.githubusercontent.com/assets/8515899/12875360/4b03cf80-cd9d-11e5-95f0-7bd3e6c43359.png)

johnaoss:
[...]

PR #1913, opened 2016-04-09
Removed dead links from issue #1802

ollyfg:


logicmoo:
Instead of just deleting links.. can we instead update them?

For example: the link  http://www.cse.iitd.ernet.in/~ssen/csl705/admin705.html

can be replaced by this link 

https://web.archive.org/web/20120430202118/http://www.cse.iitd.ernet.in/~ssen/csl705/admin705.html

vhf:
Thanks for the help, this has already been covered by another PR. :+1:

PR #1917, opened 2016-04-11
Add Xamarin.Forms C#

unseen1980:
1200-page ebook is a deep dive for C# programmers who want to write applications using a single code base that targets the three most popular mobile platforms: iOS, Android, and Windows


vhf:
Thanks! I'm not familiar with aka.ms, is it a file hosting service? An url shortener?

Is there a more authoritative source for this book? A homepage?

unseen1980:
Hi vhf,

It's microsoft's shortener. Maybe we can use this link instead?
https://developer.xamarin.com/guides/xamarin-forms/creating-mobile-apps-xamarin-forms/ 

I thought pointing directly to download link was better.

Thanks

vhf:
Yep, I'd prefer this link. I'll take care of this. Thanks!

PR #1925, opened 2016-04-22
Add notation for Ansible Up and Running Book

joelhandwell:
The book is available for first three chapters, and email registration is required to get it.


vhf:
Thanks!

PR #1935, opened 2016-05-15
previous link of Google Java Style Guide is dead

ghishadow:
the link to **Google Java Style Guide** 

> http://google-styleguide.googlecode.com/svn/trunk/javaguide.html

 is broken . after googling i found this link 

> https://google.github.io/styleguide/javaguide.html

 which means they shifted this guide from svn to github.


vhf:
Thanks!

PR #1951, opened 2016-06-11
Add New Go Book

vikashvverma:
Added URL of Free Go Book that is available in `PDF`, `EPUB` and  `MOBI` format.

URL: [https://leanpub.com/golang-tdd](https://leanpub.com/golang-tdd)


eshellman:
"book is 10% complete"

vhf:
Yep, it's a bit too light at the moment. And it seems to be abandonned.

PR #1957, opened 2016-06-24
Filtered Django resources

Shafiul-Hasan-tareq:
* Added new documentation for Django. 
* Removed http://gettingstartedwithdjango.com as the site is down for a long time.
* Updated the Tango with Django book
* Removed the "The Django Book" as it was very old version of Django


vhf:
Nice, thanks!

PR #1963, opened 2016-07-01
Create Standards for developing flexible, durable, and sustainable HT…

pridemusvaire:
…ML and CSS.

### DIFF
--- a/Standards for developing flexible, durable, and sustainable HTML and CSS.
+++ b/Standards for developing flexible, durable, and sustainable HTML and CSS.
@@ -1,0 +1,1 @@
+http://codeguide.co/


Standards for developing flexible, durable, and sustainable HTML and CSS.
```diff
@@ -0,0 +1 @@
+http://codeguide.co/
```
eshellman: nice addition, but add it to free-programming-books.md (39), not as a separate file

pridemusvaire: thanks will do that

PR #1976, opened 2016-07-05
Add openHPI to list of MOOCs

franzliedke:
Full disclosure: I work there. ;)

Still, I think it fits the list, so if you agree...


vhf:
Thanks!

PR #1990, opened 2016-07-07
Adding API Best Practices to Software Architecture

veesahni:


vhf:
Thanks!

I'm not sure if this really qualifies as a *book*.

veesahni:
Agreed that it's not packaged as a _book_.  Nonetheless, it's a comprehensive post on the topic, more so than [this other related book](http://apigee.com/about/blog/technology/announcement-free-ebook-web-api-design) that is included.  So I thought it would be a valuable resource to share.

PR #1994, opened 2016-07-07
Fix a broken link to 'Practical PostgreSQL'

jwworth:
The original link is now a 404. I'm not sure if the authors intend this resource to remain free; if not, it should be removed. If so, here's a working link.


vhf:
Thanks!

PR #2017, opened 2016-07-28
copy the makefile to 项目相关

HDelbert:


vhf:
Is there a copyleft notice somewhere showing this book is really free?

PR #2019, opened 2016-08-01
fix 404 for Google JavaScript Style Guide

tnga:
the guide's code has been switched to github. The old link was giving access to the xml source so i also do the same. However, i recommend to change it to the html page (the gold is free book, not free code!).

*so if you agree i can do the change*


vhf:
Thanks!

PR #2023, opened 2016-08-02

potatogopher:
Add EPUB, PDF, MOBI for Elixir's Getting Started tutorial


eshellman:
@vhf should probably decide on a policy for supplemental format links. adding a second line for the same book seems like overkill.

vhf:
Yep @eshellman, I think we already have something about this, here is an example of the preferred way of adding supplemental links: <https://github.com/vhf/free-programming-books/commit/b6a611c9ec83bd584c25733eeed57053294e45dd#diff-09e7ccff8516d5272cc11dcbff51c4bbR1259>

vhf:
Thanks !

PR #2032, opened 2016-08-06
Adds Phoenix section and Versioned APIs book

elvio:
* Adds Elixir > Phoenix Section
* Adds [Versioned APIs with Phoenix](http://elviovicosa.com/books/versioned-apis-with-phoenix) book

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -34,6 +34,7 @@
 * [DTrace](#dtrace)
 * [Elasticsearch](#elasticsearch)
 * [Elixir](#elixir)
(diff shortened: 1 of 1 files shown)


eshellman:
"Versioned APIs with Phoenix" requires email sign-up. Listing needs to add a note like "(PDF - need email confirmation)" to be consistent with other titles; someone (other than the author) should also confirm that the ebook is in fact available.

vhf:
@eshellman Well spotted! Ebook tested and note added.

PR #2039, opened 2016-08-13
Update free-programming-interactive-tutorials-en.md

antimatter96:
Added go by example


vhf:
Awesome resource, but not interactive. Could you add it to the `-courses` file instead?

PR #2074, opened 2016-09-10
add zh_TW (traditional chinese) md

Billy4195:
This is translate from free-programming-books-zh.md

### DIFF
--- a/free-programming-books-zh_TW.md
+++ b/free-programming-books-zh_TW.md
@@ -1,0 +1,662 @@
+## 目錄
+
+* [語言無關](#語言無關)
+  * [IDE](#IDE)
+  * [MySQL](#mysql)
+  * [NoSQL](#nosql)
+  * [PostgreSQL](#postgresql)
+  * [Web](#web)
+  * [WEB伺服器](#web伺服器)
+  * [其它](#其它)
+  * [函數式概念](#函數式概念)
+  * [分布式系统](#分布式系统)
+  * [在線教育](#在線教育)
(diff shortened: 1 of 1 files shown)


Billy4195:
Does anyone know how to fix "Alphabetical ordering" error ??

vhf:
Is this translation necessary?

`free-programming-books-zh.md` is simplified chinese, right? How many people know traditional chinese but cannot read simplified chinese?

Billy4195:
For those who are used to use traditional chinese will be convenient to read this translation.

Because there are still some difference between simplified and traditional chinese can't be translated easily.

eshellman:
@Billy4195 Easiest way to fix the ordering is to use the unicode collation of the headings- bopomofu or another chinese-native collation is not something travis will know about.

@vhf This appears to be a zh-TW transliteration of the headings, not a list of zh-TW resources. I think that breaks the pattern used for other languages; but then putting both simplified and traditional chinese resources in the same file already breaks the pattern. So I suggest to add traditional headings in parens in the zh file where the traditional characters differ from simplified. This will make a combined resource easier to maintain. (note that traditional characters are also used in Hong Kong).

vhf:
Thanks for your inputs @Billy4195 and @eshellman . I quite like the idea of adding the traditional Chinese translation of the headings in parens.

eshellman:
Should we leave this open?

eshellman:
I have created a help-requested issue #2384  to replace this PR.

PR #2085, opened 2016-09-29
Fix broken link for "Writing Robust Java. CodeJava | Japanese translation edition"

wanashi:
Current link redirects to another page and it says "our hosting service is closed on 01/31/2016"


vhf:
Thanks!

PR #2088, opened 2016-10-01
broken sql pdf link replaced

navyad:
For MySQL section:

[http://downloads.mysql.com/docs/mysql-tutorial-excerpt-5.1-en.pdf](http://downloads.mysql.com/docs/mysql-tutorial-excerpt-5.1-en.pdf)

Above broken URL replaced with following one:
[http://downloads.mysql.com/docs/mysql-tutorial-excerpt-5.5-en.pdf](http://downloads.mysql.com/docs/mysql-tutorial-excerpt-5.5-en.pdf)


vhf:
Thanks!

PR #2090, opened 2016-10-01
added head first sql

navyad:
msql new book "Head First SQL" added.


vhf:
This book is certainly not free.

PR #2093, opened 2016-10-04
Broken Link ( El maldito libro de los Descarrilados )

DeathMetalCoder:
### Description
After attempting to access the link I discovered that it is broken. Spent some time researching if the link was moved or if I could find a new link to the PDF but had no success.


vhf:
Thanks!

PR #2097, opened 2016-10-04
Added Hack Lang

thebouv:
Added Hack Documentation / Getting Started online book and categorized under PHP since it is a dialect.

I was torn on creating it as a separate language entirely or putting it under PHP.  Seems best under PHP but could be convinced otherwise.


eshellman:
Hmmm... best way to stall a PR is to express a bit of doubt. Repo maintainers wait for comment, then it gets put to the back burner, then to the second page of PRs!

I know little of PHP or Hack, but I think a separate listing is simplest. Someone looking for a Hack book will look under "H"; Hack's description gives the impression of a language distinct from PHP.

eshellman:
on the principle that no PR should be open more than a year, I'm going to close the PR and open a new one, with Hack as a language heading (not under PHP).

aqlynn:
2017年10月13日 上午2:29，"eshellman" <notifications@github.com>写道：

PR #2101, opened 2016-10-06
Add HTML version of "Using Swift with Cocoa and Objective-C"

suzannehamilton:
There is already a link to the iBook version of the book, but the HTML version is more useful if you want to read it without iTunes or an iOS device.


vhf:
Thanks!

Please use the format described at the end of [this section](https://github.com/vhf/free-programming-books/blob/master/CONTRIBUTING.md#formatting).

suzannehamilton:
I've updated the formatting for both books in the Swift section. Sorry about that!

vhf:
Thanks!

PR #2103, opened 2016-10-06
Change url to a better source

quobit:
ObeytheTestingGoat.com is the author's website for the book. I think the presentation is nicer than O'Reilly site


vhf:
Nice!

PR #2114, opened 2016-10-17
Added "CodeIgniter" Wikipedia Indonesia links

alifarianda:


thebouv:
@alifarianda Was about to comment on the Wikipedia link not being a free programming book, but looks like you removed it.

But then the book you replaced it with is in English (so does it belong here in free-programming-books-id for a particular reason?) and is for a CodeIgniter version that is beyond legacy (2.x is legacy, 1.7 is before even that).

Just some friendly feedback. Not responding in any official capacity.

eshellman:
Thanks @thebouv and @alifarianda 

Picking this up from some months ago... I agree the resource is english, and belongs in the english list not the -id list.

PR #2117, opened 2016-10-18
Add Ry's Git Tutorial

wizh:
Amazing overview of Git - both something practically applicable as well as a good overview of the internals.


vhf:
Thanks!

PR #2129, opened 2016-10-23
Add bash

kompowiec:
and site from which I learned about webdev


eshellman:
Thanks @faraco and @kompowiec. @faraco is correct about the syntax lint.

I see that the polish list includes both books and interactive tutorials. Not to hold up this PR, but would it be a good idea to separate them?

kompowiec:
i think so that tutorials is not enough to separate IMO. Apki.org actually is probably the only interactive course in polish.

PR #2130, opened 2016-10-23
update verilog

setyawanfa:
update a free source of verilog quick reference

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -2158,7 +2158,7 @@ For resources on Angular, Backbone, D3, Dojo, Ember, Express, jQuery, Knockout, 
 ### Verilog
 
 * [Verilog Tutorial](http://www.asic-world.com/verilog/veritut.html)
-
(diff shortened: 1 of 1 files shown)


free-programming-books.md
```diff
@@ -2158,7 +2158,7 @@ For resources on Angular, Backbone, D3, Dojo, Ember, Express, jQuery, Knockout,
 ### Verilog
 
 * [Verilog Tutorial](http://www.asic-world.com/verilog/veritut.html)
-
+* [Verilog Quick Reference](http://web.stanford.edu/class/ee183/handouts_win2003/VerilogQuickRef.pdf)
```
eshellman: usually we don't like to use links to websites set up for specific courses like this one because the tend to go away. A quick google search turns up many variations on this quick reference; is there a reason to use this one over any other?

eshellman:
Also, pay attention to the syntax linter error messages.

PR #2143, opened 2016-10-29
added Kattis to competitive programming and problem sets

tombiju:
Added **Kattis** (https://open.kattis.com), a great site which hosts competitive programming competitions and also has problem sets from past competitions one can work on.


vhf:
Thanks!

PR #2144, opened 2016-10-29
o-Reilly open source books / gitbook branch

devs3:
link : http://www.oreilly.com/openbook/


eshellman:
I'm not sure what's going on here. Is there a gitbooks build that I should be aware of? shouldn't the ebooks that get built be put in releases? (Look at GITenberg for examples.)

eshellman:
Since this PR is a year old, I'm closing it. I think it's an interesting idea, and I love what the gitbooks folks have been doing, but weighing the benefit against the added maintenance, I think we need to stick to our knitting. If there was a fork that maintained releases, we could link to the releases. Thanks @devs3!

PR #2164, opened 2016-11-14
add Ruby Web Dev: The Other Way

iJackUA:
this book is still WIP, but already filled with main content
has about 120 readers on Leanpub


vhf:
Thanks!

PR #2168, opened 2016-11-16
Add 'Begin Latex in minutes'

sgleung:
Added 'Begin Latex in minutes' to the 'LaTeX' section and fixed typos in 'Arbitrary Latex Reference' (Arbirtrary -> Arbitrary)


vhf:
Thanks!

PR #2169, opened 2016-11-16
Added I Love Ruby official website

mindaslab:
Added I love ruby official website so that people can get latest updated book every time. This was done because I consider dropbox as a spyware and so have moved to openmailbox for my book distribution.


vhf:
Thanks!

PR #2170, opened 2016-11-16
Added three great R books

Dripdrop12:


eshellman:
1. not clear that R Graphics Cookbook is free: http://www.cookbook-r.com/Graphs/
2. "R for Data Science" and "The caret Package" are great additions.
3. Alphabetize the entries. g after f and d before s.

Dripdrop12:
What about books like those on leanpub.com that can be $0?  Are those free?

eshellman:
read the travis report - need to remove trailing slashes.

There are lots of Leanpub books on here. Sometimes, though, the intent of the authors is to use the $0 minimum as a temporary promotional price which is hard to track.

vhf:
Thanks!

PR #2193, opened 2016-12-14
Correction of Order and Adding new Language

kreativmonkey:
# First Change Order
I Change the Order of the Language List. iOS behind H not before G!

# Second Adding Go Tutorials
Adding german versions of the Go Tutorials from the English "free-programming-books" list.


vhf:
Thanks!

PR #2194, opened 2016-12-15
remove broken link & sort lexicographically

hanefi:
- I deleted the `Windows Phone` entry in the index since there are no books under this category

- Changed the order of `R` and `Python` so it is lexicographically ordered


vhf:
Thanks!

PR #2198, opened 2016-12-22
Perl 6 : new book

faraco:
Added learn perl 6 in y minutes site. Released under


vhf:
Thanks! This is not a book though. ;)

PR #2202, opened 2016-12-30

GiggleAll:
commit one circle about PHP


vhf:
Thanks! This doesn't look like a book though.

PR #2203, opened 2016-12-30
Added From Python to NumPy

quobit:
From Python to NumPy is a beautiful edited book about the transition from using Python to NumPy through vectorization. Licence is CC BY-SA-NC


eshellman:
Nice resource.

Adaxry:
It's helpful

vhf:
Thanks!

quobit:
You're welcome!

PR #2205, opened 2017-01-01
Add and Update about Standard ML

yonta:
1. Add Standard ML slides in Japanese.
2. Update old version Standard ML document to newer one.
3. Add Standard ML web page in English.
    * This web page already is in Japanese, and also has English version.


vhf:
Thanks!

PR #2206, opened 2017-01-01
Replace "Learn Lisp the Hard Way" dead link

zdroid:
Basically, LLTHW site is dead. Luckily, there's a GitHub-repository copy of the original book, last updated 1 year ago (I hope that's recent enough).

If the book is hosted anywhere else, please write it down in the comments.


vhf:
Thanks!

PR #2210, opened 2017-01-02
add PythonToScala

intracer:
* [PythonToScala](https://wrobstory.gitbooks.io/python-to-scala/content/) - Rob Story


eshellman:
Nice resource, but you need to use alphabetic order.

PR #2218, opened 2017-01-05
Update Oracle links

tschf:
The `tahiti.oracle.com` documentation index was switched off some time ago, so updated it to `docs.oracle.com`. I also note there was no section for Oracle PL/SQL so I removed that entry from the index.


vhf:
Thanks! I'll leave this PR open because I might decide to remove this link from the file. It's not really a book, and hardly something you would find as a 1st google result. I'm not sure it adds anything to free-programming-books.

tschf:
No probs! That makes sense :)

vhf:
Awesome, thanks 👍

PR #2230, opened 2017-01-12

mbronek7:


vhf:
Thanks!

This doesn't really look like a book or some lengthy-ish content. Would you say it's a book?

PR #2236, opened 2017-01-15
Fixed link for Machine Code for Beginners

Mixolydia97:
Original linked to a page whose download links were all broken. Linked instead to publisher's website, where the book can be downloaded for free.


vhf:
Thanks!

PR #2253, opened 2017-01-31
Added markdown tutorial

Liza-S:


vhf:
I usually accept things that are not-really-books-but-close-enough in the `-books` files, but this one is a bit too far from a book. :)

PR #2257, opened 2017-02-04
Fixed A Byte of Python book entry

paulofreitas:
- Renamed book to A Byte of Python
- Fixed book ordering
- Changed link to https
- Fixed Python version (3.x)
- Added available formats


vhf:
Thanks!

PR #2261, opened 2017-02-05
Added Slax Linux Documentation

Liza-S:


vhf:
Thanks! This is a bit too far from what I consider as a book.

PR #2287, opened 2017-03-03
Delete a non free book.

u-combinator:
1. The book Realm of Racket, is not free.  
2. Added a repp of Unix system programming in OCaml


eshellman:
Agree that "Realm of Racket" is not free.
"Unix system programming in OCaml" is an excellent addition.

vhf:
Thanks!

PR #2314, opened 2017-03-27
Removing outdated books

TheBestMovieIsYourImagination:
Removing outdated Java books from Update free-programming-books-it.md


shang-vikas:
Can I add torrent link of a paid book?

ziyaddin:
> Can I add torrent link of a paid book?
No, because:
- it is torrent link of paid book and it is illegal (**the main reason**)
- this repo is for free book sources, not paid ones.

vhf:
I would prefer keeping books about older Java versions. Some people might still need to maintain software using these versions.

I would welcome a PR adding Java version info to these links though! This way the reader can quickly identify which books are suited to their need.

PR #2321, opened 2017-03-31
Fix grammar typo for Azerbaijani docs reference

ziyaddin:
There is a typo in Azerbaijani docs reference. "Azerbaijan" (country) should be changed to "Azerbaijani" (language).


vhf:
Nice, thanks!

PR #2328, opened 2017-04-04
Update Competeivie Programming Link

venkateshilango7:


free-programming-books.md
```diff
@@ -306,7 +306,7 @@
 
 #### Competitive Programming
 
-* [Competitive Programming, 1st Edition](https://sites.google.com/site/stevenhalim/) (PDF)
+* [Competitive Programming, 1st Edition](http://cpbook.net/) (PDF)
```
eshellman: since only the first volume is free, I'd recommend using https://cpbook.net/#CP1details This should also fix the travis warning

PR #2338, opened 2017-04-13
added Linux Shell Scripting Tutorial (LSST) v2.0

shivam-k:


eshellman:
Great to get the 2.0 version; is there value in retaining the 1.0x version of the same book (link just above it)?

vhf:
Thanks!

PR #2344, opened 2017-04-25
Add playground section for Haskell

lerthe61:
Try Haskell https://www.tryhaskell.org/


lerthe61:
Fix everything related to mardown formatting rules

PR #2346, opened 2017-04-26
Add "Design With FontForge" (CC licensed)

eshellman:
There are also French and Chinese translations which can be added but the misc category doesn't exist in those lists.

Font design is on the boundary between design and programming. Hope there's room for this.


vhf:
Thanks!

PR #2352, opened 2017-05-10
添加 Ruby on Rails Guide

zcq100:
添加 Ruby on Rails 简体中文指南


eshellman:
Great addition! We'd really appreciate it if you made the linter happy. It wants the lists in unicode sort order (笨 after R).

PR #2360, opened 2017-05-21
Add free eBook: The Road to learn React

rwieruch:
The book is perceived as open source book to learn React. People like and share it in the community. I would love to see it in the list of free React books, even though people can pay what they want to support events like this: https://www.robinwieruch.de/giving-back-by-learning-react/


eshellman:
Thanks!

PR #2373, opened 2017-06-02
Added Solidity programming language and book

alcole:


eshellman:
This resource describes itself as an article. Is it the best free book-like resource on Solidity? Also, it's 2 years old. Is it still useful?

Also I note that we don't list the book it cites as the best place to start, https://github.com/bitcoinbook/bitcoinbook

alcole:
Both good points. 
I found the book / article when searching for books, so perhaps it doesn't fulfill the strict definition, but there is not much out there (free or otherwise).
On your second point, the book is on Solidity which is a programming language for programming smart contracts on the Ethereum block chain. The Bitcoin book that is linked to is more about the bitcoin / blockchain environment, rather than programming smart contracts, but it might be of interest here, I'll check if it is added.

eshellman:
On the Solidity article, I agree it could be a book, and it's in a free-book desert, so we should add it. The bitcoin book is a great addition. Could you add the authors in the listings for these two books? TIA.

alcole:
Done! cheers

eshellman:
much appreciated!

Issue #2383, opened 2017-06-08
Let's start adding author names where appropriate


eshellman:
* many programming books have names that can be confused for similar books: e.g.  "introduction to Java". It's very helpful to browsers to see the author names
* The identity of books with translated titles may not be apparent
* People googling for books will more easily find our lists
* (perhaps most important) Authors who make their books available for free deserve recognition

If you'd like to assist in this effort: 
* submit PRs for manageable sections of the lists. (10 or so books at a time.)
* if there are a lot of authors, use "First Author et al."
* sometimes there's no useful author, don't sweat it.
* reviewing other contributors' PRs is almost as helpful to the admins as submitting new ones.

tedyoung:
Would it be useful to add the year as well?

eshellman:
A previous discussion #1926 (which I see @tedyoung has found) suggested that adding the year for older books would be a good idea.

Issue #1926, opened 2016-04-22
Year of publishing


aminought:
I'm not sure if anyone have suggested it already, but what do you think about signing of a year of publishing near with a title of a book?

vhf:
Thanks for the idea. There are pros and cons.

I think the cons outweigh the pros:

* years aren't always good indicators of the current value of the information contained in a book
* would require a lot of work
* adds some overhead to the information the lists already contain

aminought:
I think you are wrong. For me and, I'm sure, not only for me, year is a very good indicator of the information actuality. If I see a book published in 2003 and a book published in 2015, there is a big probability that I'll choose the second one because it contains more modern information.

vhf:
I'm open to discussion on this topic, I'd be glad to take other opinions.

eshellman:
One issue with tracking years is that many resources in the repo are being periodically updated. for the purposes of the repo, this would create more maintenance, more PR's to keep the info current. Side databases such as the one @vhf has worked on may be a better means of providing this info; I'm working an a git-tag-triggered metadata update scheme that might work, too.

Having said that, it would be useful, and perhaps not so much work, to add publication years only for resources that are older than ~5 years (~10 years?).

brutalcrozt:
I am agree with @amninought, but @vhf also had good point. If we only measure the book by it's publication year, it will not make sense.

> would require a lot of work.

Then leave it to the community via PR. Encourage to submit a PR, even just for one book. Personally I feel no such things is literally free. It will cost time or will mostly. It's a return value as they found the list worthy.

> adds some overhead to the information the lists already contain
I am disagree with this, we know information now is cheap. What is expensive is understanding.

my point : it's worth add one more identifiers. Because surely it help give identifiers to suit the consumer needs :smile: . It's just like add one more optional column on database, a mother name. Reality shown not all child know their mother name.

borgified:
could be an option if we could come up with a programmatic way to determine this. i mean it doesnt have to be perfect but if there was a way to automate for perhaps 50% of the cases with good accuracy then that might be a good starting point

michaeljota:
@aminought I think that's exactly the point that @vhf means, maybe the older book it is better explaining or have better examples. It's old, what so ever, but for majority of software languages some old ways to do things still being right, even if they are not, it doesn't mean the book is worst compared to the newer.

Still, I agree that using the year of publish it's a good idea, even for those whom are currently in development.

[...]

PR #2387, opened 2017-06-10
regularize pub year format

eshellman:
and add to guidelines

supports #1926 and #2383


eshellman:
@tedyoung would appreciate comments

marlazinger200:
1

PR #2471, opened 2017-06-22
Added "High level async programming with Rx"

petroemil:


eshellman:
No need to be modest, please add the author name!

It's not required for F-P-B, but it's useful in general to explicitly add a LICENSE file to the book's repo.

petroemil:
Thanks for the feedback. I updated the listing to include my name.
I wouldn't necessarily insist on having a LICENSE file, but if you could provide me a sample, I'd include it.

eshellman:
This looks  really useful, thanks! Take a look at [this repo's license](/LICENSE) for a really simple example.

PR #2472, opened 2017-06-25
添加了一本mysql书籍(add a book about mysql)

anxiaokai:
添加了一本mysql书籍(add a book about mysql)


eshellman:
This doesn't appear to be an authorized copy. Do you have any information that it is?

PR #2475, opened 2017-06-28
Modified Railscasts, now Free

drewsmith:
Ryan Bates of Railscasts recently announced that all screencasts are free: http://railscasts.com/announcements/13


eshellman:
good update!

PR #2478, opened 2017-06-29
added webpack bin, js bin playgrounds

drewsmith:


eshellman:
This is the first I've looked at the playground list. Do you think it's organized well? I see some duplicate listings.

PR #2480, opened 2017-07-01
Remove hardrockcoffeescript.org

ctrlaltf2:
It looks like this domain has been registered to a different owner.


eshellman:
If you think it's worth keeping, it's at http://alchaplinsky.pro/hard-rock-coffeescript/

Issue #2486, opened 2017-07-04
Bootstrap links under HTML are poinitng to a login page


saurabh-pandey:
Hi,
First of all thanks a lot for assembling such a great and expansive list. I was helped a lot from the links provided.
While going through some new topics I was exploring, I found that the Bootstrap links (both of them) under HTML were pointing to the same login page.
I have attempted a fix for them in my local branch and would raise a Pull Request asap.
Thanks,
Saurabh

saurabh-pandey:
Hi,
Have raised a pull request (#2487). Can you please check and provide your suggestion(s) ?
Thanks,
Saurabh

eshellman:
worth checking if we have any other links to this site.

saurabh-pandey:
Sure. Will update if I find something. Thanks.

PR #2487, opened 2017-07-04
Fix links of bootstrap under HTML

saurabh-pandey:
The fix ( for Issue #2486 ) I have attempted is that now instead of pointing to the login page, the links now take you to the respective Download PDF page.


eshellman:
Nice improvement!

saurabh-pandey:
Thanks a lot. This was my first on github !!

PR #2489, opened 2017-07-06
Link to free React course

hrishimittal:


eshellman:
looks good. Would you double-check that the category organization is consistent with other list files? Because we also have react in javascript frameworks. (Not necessarily a problem, but I don't have time to check myself at the moment.)

hrishimittal:
@eshellman I considered adding it to the JavaScript frameworks resources list, but I thought maybe it's not right because it's a course. However, now I see there some courses on that list, so I would like to put mine there too. If that's ok, I'll push a change.

hrishimittal:
@eshellman I've moved the link to the JavaScript frameworks resources list. Please check and merge.

eshellman:
I'll go back and try to understand why the javascript frameworks list has both books and courses; it's confusing to both of us. But better to raise the question as a separate issue. Thanks for your help and your understanding!

PR #2507, opened 2017-08-07
added a django ebook

matikin9:
wasn't completely sure how to format the django version, considering the book author is in the process of updating chapters from v1.8 to 1.11.


eshellman:
check back in  6 mo, see if it still looks right. Thanks!

PR #2516, opened 2017-08-29
1. Domain name http://reactjs.cn was sold and it is not related to re…

nivort:
…act js for now. So I change it to another available site.

2. Tweak React to React.js


eshellman:
thanks for the fix

Issue #2353, opened 2017-05-10
Formatting for books with many versions


eshellman:
Many books, especially manuals, have an edition for every version of the software. Some of our lists are presenting one entry per version(for example PR #2350 ), others are only listing the latest version, sometimes a single link for multiple versions is presented.

What's the best way for these to be listed? When is a version so old we should omit it? Should the the practice be different for different language lists?

scraggo:
I agree, I think they should be visually grouped. Here's a thought: sublists, most recent book first. One example:

* How to Think Like a Computer Scientist: Learning with Python - Allen Downey et al - [MULTIPLE EDITIONS]
    * [How to Think Like a Computer Scientist: Learning with Python, Interactive Edition](http://interactivepython.org/courselib/static/thinkcspy/index.html) (3.2)
    * [How to Think Like a Computer Scientist: Learning with Python](http://www.greenteapress.com/thinkpython/thinkCSpy/) - Allen B. Downey, Jeff Elkner and Chris Meyers (2.4)
    * [Think Python](http://www.greenteapress.com/thinkpython/) - Allen B. Downey (2.x & 3.0)

eshellman:
that seems reasonable. Could you take a look at what the linter says about that? https://github.com/vhf/free-programming-books-lint

scraggo:
I changed my mind, making the latest version the toplist element and the previous versions the sublist elements. I'll do a PR and you can tell me what you think. The linter didn't find any issues :)

eshellman:
Thanks everyone who contributed, the suggestions are now implemented!

tunggnu:
I have found an issue. When we have a title with more than 3 editions, we can't list them in descending chronological order, due to the rule "titles must be in alphabetical order".

For example, in the #5299 PR, my first commit could not pass Travis CI because it contains:

* [How to Think Like a Computer Scientist: Learning with Python, Interactive Edition](https://runestone.academy/runestone/books/published/thinkcspy/index.html) - Brad Miller, David Ranum, Jeffrey Elkner, Peter Wentworth, Allen B. Downey, Chris Meyers, and Dario Mitchell (3.2)
[...]

PR #2517, opened 2017-08-31
Python books: Formatting for multiple editions #2353

scraggo:
The latest version is the toplist element and the previous versions are the sublist elements.


eshellman:
I like this. Since it's a new thing, I'm going to wait a bit before merging it, so that others have a chance to propose options or point out problems. Thank you!

eshellman:
It would be nice to add some documentation about this in CONTRIBUTING

PR #2526, opened 2017-09-05
Remove broken link

emre-yilmaz:
This site not active.


eshellman:
Thanks for the PR. Also. I think other links on this page may also need looking at.

PR #2532, opened 2017-09-08
Add link for pdf and zipped html for a C book

manaskarekar:
Include downloads link for a zipped html and pdf versions of a C book.


eshellman:
Excellent!

PR #2533, opened 2017-09-09
Update README.md

karanbhangui:
Updating original credit as per: https://news.ycombinator.com/item?id=14776982

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [View the English list](/free-programming-books.md)
 
 ## Intro
-This list was originally a clone of [stackoverflow - List of Freely Available Programming Books](http://web.archive.org/web/20130824154208/http://stackoverflow.com/a/392926) by George Stocker. 
+This list was originally a clone of [stackoverflow - List of Freely Available Programming Books](http://web.archive.org/web/20130824154208/http://stackoverflow.com/a/392926) by Karan Bhangui. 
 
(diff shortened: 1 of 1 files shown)


eshellman:
Thanks for showing up! So the reason @gortok was credited here by @vhf was because he (@gortok) posted the list from the wiki onto StackExchange, and thus incidentally got his name attached to it? That's kind of cool, too; maybe you could flesh out the history here as told in the hacker news comment.

vhf:
Hey! I don't have access to the full revision/history of the [original StackOverflow](https://stackoverflow.com/questions/194812/list-of-freely-available-programming-books/392926) ([archive](https://web.archive.org/web/20140606191453/http://stackoverflow.com/questions/194812/list-of-freely-available-programming-books/392926)) post but I credited @gortok because the list could be found in his answer.

If I understand the story correctly, @karanbhangui asked the initial question, @gortok answered with a list and kept it up to date until @gortok's answer became a community/wiki post on SO.

karanbhangui:
Happy to keep credit as is.I'm just glad to see how much this has flourished over the years! I remember starting my first semester of engineering in Sept 2008 and hating how expensive and mediocre many of the textbooks we had to buy were for subjects that hadn't changed in a long time. I think the adoption of open source textbooks has taken off over the decade since, but a long way to go.

PR #2548, opened 2017-09-27
This book should be in Mathematics, not Mathematica

iggy14750:
I looked through this a little bit, and I can't find any Mathematica code at all; it appears to just be a math book.


eshellman:
Thanks for the correction!

kincaider:
59e1c0d

PR #2566, opened 2017-10-07
Added Kubernetes

rasor:


eshellman:
I'm hesitant to add promotional free chapters to this list, especially when users are asked for email addresses. I'll leave this PR open for a week or so in case others want to make a case for Kubernetes as an exception.

PR #2570, opened 2017-10-08
Revoming a book which is not free at all

pirrmann:
I'm sure @tpetricek can confirm his book is not free


eshellman:
Agreed!

eshellman:
Github has an fair, open and easy to use DMCA removal process should @tpetricek or Manning elect to use it.

tpetricek:
Thanks for removing the link. It's worth noting that the link also points to an early draft with typos and very ugly formatting. So buying the final ebook not only gives you a good feeling of supporting the author (who spent some two years working on it), but you'll also get a much nicer version of the book with fewer issues 👍

thokaka92:
_Sent from my P10 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

thokaka92:
_Sent from my P10 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_thokaka92

PR #2571, opened 2017-10-09

YouTuQiong:
Add Django 1.8.2 Chinese document for Python


eshellman:
Great to see the django translation complete! Note that we've added a guideline about "in progress" works to CONTRIBUTING that might apply to other translations at usyiyi.cn

PR #2575, opened 2017-10-11
Add Mercurial as a category and add a Mercurial Course.

TurekBot:
I'm not sure if the hashtag works. Any help there?


eshellman:
Works for me! Thanks!

PR #2576, opened 2017-10-12
Add Python courses

tonkytonky:
Add some Python courses on Russian. After approval I'd like to add some more. 
This is my first pull request, so I hope I'm doing all the things correctly 😄

PR #2577, opened 2017-10-12
reorder russian courses

eshellman:
and add link to readme


eshellman:
#2576 was merged with failing linter, perhaps because it added a new list

PR #2581, opened 2017-10-14
R: Add Introduction to Statistical Learning by Hastie et al

pavopax:
Got to have this classic by Hastie, Tibshirani et al


eshellman:
great addition! but need to put the list in alphabetical order; our linter is not so fancy that it ignores "The " or "A " at the beginning of titles.

Issue #2151, opened 2016-11-04
New Verilog Book(?)


WanderAway:
http://www2.dc.ufscar.br/~marcondes/netfpga/FPGAPrototypingByVerilogExamples.pdf

FPGA prototyping by verilog examples. while not strictly speaking a verilog tutorial, it is pretty helpful. Wondering if it's appropriate to add to the list. 

and this:
http://vol.verilog.com/VOL/main.htm

Thought I might contribute since the verilog section seemed a bit empty

mramdas:
Yes, I think these would be useful.  There are lot of students/engineers looking forward to resources like this.
I would also like to add following  FREE  SystemVerilog online course to the free course list
http://verificationexcellence.in/online-courses/soc-verification-using-systemverilog/

eshellman:
the first link from @WanderAway does not appear to be free. The second looks good.

The link from @mramdas looks appropriate too. PR's welcome.

eshellman:
fixed, thanks to @charlottetan

PR #2587, opened 2017-10-16
Add/verilog courses #2151 

charlottetan:
Renamed `SystemVerilog` section to `Verilog / VHDL / SystemVerilog` because they are 3 different hardware description languages. Grouped them together so that all the Verilog resources can be found in one spot. Added new course as suggested in #2151


eshellman:
This is really helpful! Much appreciated.

PR #2593, opened 2017-10-17
Add Clojure by Example

javbit:
It's an example-based alternative to reading a long, in-depth book.


eshellman:
That's a nice site! Could you add the author name? I think it's especially useful when the title is something generic like "Clojure by Example". See #2383

javbit:
Thanks! Yeah I'll get on that.

eshellman:
Thanks!

PR #2595, opened 2017-10-17
Update README.md

asgs:
Minor Markdown fixes.


eshellman:
There's certainly much that can be done to improve the README. I agree this is an improvement; it just raises the question of what else we might do. Thanks!

PR #2599, opened 2017-10-17
Add Ecto section and getting started link to Elixir

zwhitten:


eshellman:
I'm not familiar with the Elixir ecosystem. Are there going to be a bunch of hex packages, each with only one entry, or does it make sense to just list them all under Elixir?

zwhitten:
@eshellman It could probably be done either way.  Ecto is one of the biggest frameworks for accessing/modeling databases in Elixir. I could move it underneath Elixir and down the road if other entries are added it could be moved back out to it's own section. Initially I didn't do it that way because it's not technically "core" elixir.

PR #2610, opened 2017-10-22
Delete Book which not free now

bestony:
the book 《Ruby on Rails Tutorial 原书第 2 版》 is not free now , So , delete it .


eshellman:
Thanks for this correction!

PR #2611, opened 2017-10-23
Fix typo list item on courses

marcolussetti:
Fix a small mistake that causes errors in displaying the courses list.


eshellman:
good catch!

PR #2620, opened 2017-10-26
Updated Berkeley CS course links. 

rcass:
Found that the Berkeley CS links were all dead (the youtube video playlist had been taken down).

I checked each of the courses for posted video alternatives. Some have links, other who didn't have alternative links were just removed.


eshellman:
Seems a shame. I wonder why they were taken down? Thanks for your efforts on this.

PR #2621, opened 2017-10-28
New Git book: DZone Refcard: Git Patterns and Anti-Patterns

gcsgpp:


eshellman:
This is an account-required resource. We accept a small number of account-required resources in the book list for important resources. Is this book important enough to outweigh the intrusiveness  of requiring an account?

DZone is already in the meta-list list.

eshellman:
Closed based on inactivity and current policy on account-required books.

PR #2623, opened 2017-10-28
Add Procedural Content Generation in Games

anadrome:
Free version of a textbook on Procedural Content Generation in Games. Seems to fit best in the Language Agnostic / Misc category, since there isn't an artificial intelligence category (this book is AI, but not primarily Machine Learning).

Full disclosure: I'm one of the authors. But I do think it fits the theme of the list. The final published version isn't free (copyrighted by Springer), but our publication agreement gives us the right to publish free PDFs of the final manuscript on an authors' website, which is what we've done here.


eshellman:
Thanks for the contribution, and thanks for getting that publication agreement!

PR #2629, opened 2017-10-31
Add "Python-Kurs"

ikarulus:


eshellman:
this is on the borderline between book and course, I'd say, but the website wants us to think of it as a course, so I agree with this placement. Thanks so much for your efforts. If you could take a look at the _de book list to see if any other resources should be moved, that would be great.

PR #2631, opened 2017-11-01
Clean up the Oberon books

anadrome:
The Project Oberon book was duplicated under both Operating Systems and Misc; plus a better link for Programming in Oberon.


eshellman:
the link you've added is a duplicate of the listing just above it. (2 books listed on that page) Perhaps the thing to do would be to use the links that go directly to the PDFs. Also, it would help to add the author(s). Thanks for helping out!

anadrome:
Good points, updated accordingly.

eshellman:
Excellent!

PR #2632, opened 2017-11-01
Add R-Fiddle to playgrounds

NicoleSingular:
Add R section to playground and add R-Fiddle to it.


eshellman:
Great addition, thanks!

PR #2634, opened 2017-11-02
Remove broken page

platinumthinker:
Author deleted page partially.
He write message on the page: This book has been removed as it will shortly be published by APress.


eshellman:
Thanks!

PR #2641, opened 2017-11-11
Update javascript-frameworks-resources.md

DenisVuyka:
- Add "Angular" section (not AngularJS)
- Add "Developing with Angular" book reference

### DIFF
--- a/javascript-frameworks-resources.md
+++ b/javascript-frameworks-resources.md
@@ -27,7 +27,7 @@
 
 ## Angular
 
-* [Developing with Angular](https://leanpub.com/developing-with-angular/) *(Leanpub, by Denys Vuika, work in progress)*
+* [Developing with Angular](https://leanpub.com/developing-with-angular/) - Denys Vuika (Leanpub, work in progress)
 
 
 ## Aurelia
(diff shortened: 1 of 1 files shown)


eshellman:
Looks like an interesting book. Just a few comments:
1. Is it intended that this book remain free once it's completed? Most of the Leanpub books F-P-B lists are at least free (meaning you can read them online without a Leanpub account) in the HTML version. My understanding is that books like this include many that do well on pay-what-you-like sales.
2. If you want to split Angular and AngularJS, you should move other Angular books to the new category.
3. I see a lot of variation in the listing of leanpub books. I prefer author name not in parentheses. What is your preference?

DenisVuyka:
Hello @eshellman, thanks for review and feedback. Please see answers below.

1. The book has pay-what-you-like model starting with the free option. It should remain free once completed.
2. Given that Angular (version 2+) and AngularJS (version 1.x) should be considered as two completely different frameworks I think it is absolutely correct putting the Angular content into a separate category. I don't see any Angular (2-5) content in the AngularJS section right now, but this can be improved in the future I believe.
3. Corrected the author name ( was just blindly copying the existing style for Leanpub related entry from the AngularJS list)

eshellman:
Thanks for this contribution!

PR #2376, opened 2017-06-05
Old site has stoped updating

117028813:
new content in http://d3.decembercafe.org/

PR #2653, opened 2017-11-24
fix zh error 

eshellman:
wrong type of whitespace. Thanks @117028813 for noticing error in #2376


eshellman:
problem caused by using \xA0 instaed of \x20 for whitespace

PR #2658, opened 2017-11-27
Added interactive OpenGL Shading Language Tutorial

nightttide:


eshellman:
That's a really cool book! I'd put it in the book section with other graphics books so that people will find it. Although it's interactive, it's interactive in the way that Jupyter-notebook authored books are, and it uses "book" in its title.
Note that the linter is picky about added line spaces, and wants the url to be "https://thebookofshaders.com"  (no ending slash).

PR #2659, opened 2017-11-28
Add Pawn language docs

Yousha:
Added Pawn language docs and Corrected 'PERL' case

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -135,6 +135,7 @@
 * [OpenMP](#openmp)
 * [OpenResty](#openresty)
 * [OpenSCAD](#openscad)
+* [Pawn](#pawn)
 * [PC-BSD](#pc-bsd)
 * [Perl](#perl)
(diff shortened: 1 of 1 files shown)


eshellman:
Interesting addition!
+ Why do you think Perl should be PERL?
+ The raw link would save a click https://github.com/compuphase/pawn/raw/master/doc/Pawn_Language_Guide.pdf
+ our linter is picky about spaces, and wants to see an indication when the target of the link is a pdf file

Yousha:
> our linter is picky about spaces, and wants to see an indication when the target of the link is a pdf file

What do you mean? I don't get it

eshellman:
1. the linter is run by Travis- we don't merge anything unless it passes the tests that Travis runs and gives a green check to.
2. the linter wants an empty space between a heading and a list, and 2 empty spaces after the list.
3. if a link url ends with ".pdf" the linter wants to see "(PDF)" after the listing.

PR #2662, opened 2017-12-01
Koma-Script

maieul:


eshellman:
would this be easer to find if it were in a TeX/LaTeX category?

maieul:
I close the PR i would submit a enw one with no problem of intgeration
@eshellman : it is already in TeX/LaTeX category, I have juste created a new section

PR #2667, opened 2017-12-21
Fix tests

170296:
deleted


170296:
[введение.docx](https://github.com/EbookFoundation/free-programming-books/files/1579622/default.docx)

PR #2674, opened 2017-12-31
remove the link to "Operating Systems Concept 9th edition"

oltolm:
It's not a book, but a bunch of slides.


eshellman:
I agree. Thanks!

PR #2675, opened 2017-12-31
Replace the link for the book "Programming in Oberon".

oltolm:
It pointed to a tutorial and not to the book.


eshellman:
I'll accept this but maybe https://www.inf.ethz.ch/personal/wirth/ProgInOberon2004.pdf is valuable too?

PR #2679, opened 2017-12-31
Added The Definitive guide to Django book

lavishsaluja:
Added one more resource to learn Django


eshellman:
I'm afraid this is not an authorized distribution of the book. Reopen if you have information that the book is really free.

PR #2686, opened 2018-01-12
Please add a new Python Book

H3li0:
I found this new book just released in January 2018.
It is a compilation from StackOverflow documentation and is freely available.

Thanks.


ShipsWithCannons:
There's a leading space before the `http` which shouldn't be there.

eshellman:
And the lists need to be in alphabetical order. Gee I wish we had a robot that would help correct these things.

eshellman:
Thanks @H3li0 for the submission, and for persevering. Thanks @ShipsWithCannons for assistance.

eshellman:
pushed the wrong button!

PR #2689, opened 2018-01-19
For people looking at the file listing for the English list

Rossolson:
This redirect may eventually swap contents with the primary /free-programming-books.md to put English on an equal footing with the other languages.

### DIFF
--- a/free-programming-books-en.md
+++ b/free-programming-books-en.md
@@ -5,4 +5,4 @@
 
 ### All
 
-* [English]: See (/free-programming-books.md)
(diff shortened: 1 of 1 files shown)


eshellman:
nice idea, but the link doesn't work for me. Also maybe some words to say "the list of books in english is [here](address) for historical reasons".

Rossolson:
@eshellman The link has been fixed and I've added the requested commentary to the line as well. I hope this is helpful.

eshellman:
OK, we'll see if folks find the helpful.

PR #2690, opened 2018-01-20
Added Angular and Java Screencasts

juniortarcisio:
Added some great free Portuguese screen-casts courses which I've used in past and some Brazilians have been using.


eshellman:
Thanks!

PR #2692, opened 2018-01-21
Uniform treatment of Leanpub books

eshellman:
free Leanpub books come in two flavors
- free to read online without an account or email address
- free to read after a "purchase" with a minimum price of "Free!" A
valid email is requested to create an account, but you can read   or
download even if the email is bogus.

After some consideration, I'm proposing that the free-to-read books
should be without any annotation, and the free-after-"purchase" books
be denoted as *(Leanpub Account or valid email requested)* . This PR
implements that for the english lists.

Comments requested!

If folks like this change, we can also change the current notations for O'Reilly and Pakt.

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -460,7 +460,7 @@
 
 * [2016 European Software Development Salary Survey](http://www.oreilly.com/programming/free/2016-european-software-development-salary-survey.csp) *(Just fill the fields with any values)*
 * [2016 Software Development Salary Survey](http://www.oreilly.com/programming/free/2016-software-development-salary-survey-report.csp) *(Just fill the fields with any values)*
-* [97 Things Every Programmer Should Know - Extended](https://leanpub.com/97-Things-Every-Programmer-Should-Know-Extended)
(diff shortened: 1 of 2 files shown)


ShipsWithCannons:
As long as this only gets done for major series, I don't see a problem.

It gets problematic if this will spawn 30 different account notices, in which case 'Account required' should suffice for all of them.

borisd:
@eshellman considering the books are free (especially given that "fake" address can be easily used and book obtained in one simple flow). Would this warning not deter people from get the books? 

What would be the purpose of this tag?

eshellman:
For context, the O'Reilly, Bookboon and Packt books (and a few others) have had the notation  *(Just fill the fields with any values)* 

@ShipsWithCannons yes, we'd like to minimize the number of different notices. "*(account required)*" is inaccurate, as @borisd noted on #2688. but without notation, people (like me!) will think that the ebook "costs" an email address. I like Leanpub and am happy to give them my email address, but if you've never heard of another ebook site, it's very reasonable to be hesitant to give them a real email address.

PR #2696, opened 2018-01-27
add "Using Concurrency and Parallelism Effectively"

oltolm:


eshellman:
Any reason for the change case on a few headings?

Alphabetization error - u comes after t

oltolm:
Yes, "Markdown" is the proper spelling.

Fixed.

eshellman:
Thanks!

PR #2697, opened 2018-01-28
remove course link

oltolm:


eshellman:
At first I wondered why this should be removed, it seems like a perfectly good course. Then I noticed: " Oh! it's in the books list, not the courses list, where it belongs."

Thanks for catching the miscategorization. I'd be happy if this was added to the courses list

PR #2698, opened 2018-01-28
Remove "JAAS in Action"

oltolm:
It's just the first chapter of the book and not a complete book.


eshellman:
we implemented the 🚧 annotation to mark books like this; this one seems to have no progress or work in a year, so I agree it should get dropped. Oh well.

PR #2699, opened 2018-01-28
Remove "What's New in Java 8"

oltolm:
Book is not free. It's a minimum of $4 https://leanpub.com/whatsnewinjava8 .


eshellman:
yep, that's not free.

Issue #2503, opened 2017-08-07
help requested sorting zh


eshellman:
#2498 could use some help. volunteers would be appreciated. 

this is a common issue, and there are subtleties:

https://stackoverflow.com/questions/1097908/how-do-i-sort-unicode-strings-alphabetically-in-python

our linter uses some javascript to decide what is "alphabetical"

https://github.com/vhf/remark-lint-alphabetize-lists/blob/master/lib/alphabetical-list-items.js

liluoao:
Thank you!

eshellman:
If any zh-speaker can volunteer to help other zh-contributors with sorting issues, it would be a great help.

zhongjiajie:
I can help sometime if you wish, but some `-zh` list sorted is odd for me. For example, some book's name in chinese start at `a` > `z` like [this](https://github.com/EbookFoundation/free-programming-books/blob/master/free-programming-books-zh.md#项目相关), others `z` > `a` like [this](https://github.com/EbookFoundation/free-programming-books/blob/master/free-programming-books-zh.md#python)

zhongjiajie:
In my  latest comment, I don't see the link 
https://github.com/vhf/remark-lint-alphabetize-lists/blob/master/lib/alphabetical-list-items.js  
to decide what is "alphabetical". Now I kown how to do it, I will help chinese's contributor if them miss some sorted issue.

eshellman:
stale

PR #2701, opened 2018-02-01
Add 4 books

brannx:


eshellman:
These are all excellent additions, thank you. 

The new entries need to be put into "alphabetical" order, which means cyrillic after roman. see issue #2503 for a similar issue; we can put out a call for help if you need it.

eshellman:
Thanks for working on it, let us know if you need help.

PR #2703, opened 2018-02-02
remove non-free books

oltolm:
"Why programmers work at night" is not free and is missing the link "Free to Read online".
"Practicing Domain-Driven Design" is only part 1 of 4 and for the rest you have to buy the book.

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -624,7 +624,6 @@ Kerridge (PDF) (Just fill the fields with any values)
 * [How to Do What You Love & Earn What You’re Worth as a Programmer](https://leanpub.com/dowhatyoulove/read) - Reginald Braithwaite
 * [Professional Software Development For Students](http://mixmastamyk.bitbucket.org/pro_soft_dev/) - Mike G. Miller
 * [What I've Learned From Failure](https://leanpub.com/shippingsoftware/read) - Reginald Braithwaite
-* [Why programmers work at night](https://leanpub.com/nightowls/read) Swizec Teller
 
 
 #### Programming Paradigms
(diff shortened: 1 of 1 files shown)


free-programming-books.md
```diff
@@ -687,7 +686,6 @@ Kerridge (PDF) (Just fill the fields with any values)
 * [Naked objects](http://downloads.nakedobjects.net/resources/Pawson%20thesis.pdf) - Richard Pawson (PDF)
 * [OAuth - The Big Picture](https://pages.apigee.com/oauth-big-picture-ebook.html)
 * [Object-Oriented Reengineering Patterns](http://scg.unibe.ch/download/oorp/) - S. Demeyer, S. Ducasse and O. Nierstrasz
-* [Practicing Domain-Driven Design](https://leanpub.com/Practicing-DDD) - Scott Millett  *(Leanpub account or valid email requested)*
```
eshellman: This book has a $0 minimum. Do you think the notation should be different? see #2692

free-programming-books.md
```diff
@@ -624,7 +624,6 @@ Kerridge (PDF) (Just fill the fields with any values)
 * [How to Do What You Love & Earn What You’re Worth as a Programmer](https://leanpub.com/dowhatyoulove/read) - Reginald Braithwaite
 * [Professional Software Development For Students](http://mixmastamyk.bitbucket.org/pro_soft_dev/) - Mike G. Miller
 * [What I've Learned From Failure](https://leanpub.com/shippingsoftware/read) - Reginald Braithwaite
-* [Why programmers work at night](https://leanpub.com/nightowls/read) Swizec Teller
```
eshellman: This book is all (well, the 70% that's written) readable at the link, has a CC license, and has a github version. So it's free to read. Do you see something different?

oltolm: Didn't know about the CC license.

free-programming-books.md
oltolm: It is only part 1 of 4 and it's not a work in progress. We have already removed incomplete books that were not a work a progress.

eshellman: At 160 pages, it seems substantial enough to be a book on its own. 

I suppose the key question is whether people wanting to learn about Domain-Driven Design would find these 10 free chapters useful or not. Perhaps folks with expertise in this area can weigh in.

oltolm:
As expected, nobody replied. I still suggest removing both books, simply because they are incomplete and the last update was a very long time ago and we have already removed books like this.

* "Practicing Domain-Driven Design" is incomplete and was last updated on 2014-01-14.
* "Why programmers work at night" is incomplete and was last updated on 2013-10-21.

Should they ever become finished, we can add them again.

eshellman:
maybe use blame to see if the people who added the books still think they should be in the list.

oltolm:
Sure, why not, feel free to ask them.

eshellman:
what do you think, @alexanderfefelov ?

eshellman:
or you, @rodrigo-medeiros ?

rodrigo-medeiros:
@eshellman it's fine for me to remove those books.

eshellman:
so as not to leave this open, I'm going to remove "Why programmers work at night" and keep "Practicing Domain-Driven Design" with a notation.

Issue #2672, opened 2017-12-28
We need a nit-fixing-bot


eshellman:
At least half of all contributions to free-programming-books  get tripped up by the linter, which is picky about blank spaces, list collation, pdf links and the like. These are things that could be fixed by a bot, which could add commits fixing minor errors. This would cut the administration overhead significantly and would reduce contributor frustration, improving the experience for all involved.

borgified:
i googled a bit and found this: https://stickler-ci.com/
i think we should be able to integrate it to our repo but first we should test it out and make sure it fits our needs. ill get around to doing this eventually but if anyone would like to give it a shot and see if it does what it says on the tin, please feel free to do so (and report back the results here)

eshellman:
Interesting!

borgified:
i still intend to revisit this soon™ but i came across this while looking for something related which i think is relevant so im leaving it here for later https://github.com/bluzi/travis-buddy

eshellman:
Maybe someone can try out  @travisbuddy on a fork and report back.

borgified:
i just tried it in my repo. you can check out the behavior here: https://github.com/borgified/free-programming-books/pull/2
if you like that, ill create a PR and then we can merge it. squash commits before merging though, no need to pollute the git history with my testing.

zhongjiajie:
@borgified I don't think it behavior achive our purpose. In your ofrk PR https://github.com/borgified/free-programming-books/pull/2  It just tell contributor whether PR failed or not, and where failed.
we want a bot to tell contributor how to pass CI, or even auto change contribute PR to pass CI test.
Am I right @eshellman

LincePotiguara:
I looked at the tool to see if it is capable of checking what went  wrong in a Pull Rrequest (PR)
----
Here is what I've found:
- Travis-buddy will check the build log and send back just what didn't succeed
- It has some default checking
- It is possible to add custom rules

[The PR](https://github.com/borgified/free-programming-books/pull/2) created by @borgified show the tool in it's "raw" form.
In order to present nicely the test fails it require configuration. In our case, we would need a script specify the rules. 
The best aplication I could find is [this](https://github.com/OsProgramadores/op-website-hugo/blob/master/travis/checkparticipantes.go) with [this](https://github.com/OsProgramadores/op-website-hugo/blob/master/.travis.yml) configuration. 
[...]

PR #2707, opened 2018-02-05

brannx:


eshellman:
This is a good example of why #2672 we need a nit-fixing bot!

eshellman:
Thanks for pushing this though! A great contribution to the Russian list.

PR #2709, opened 2018-02-05
Added  Edx Machine Learning course by Microsoft

prateekkalra:


eshellman:
The link you propose doesn't let you see anything if you're not enrolled. Is it the same as this:
https://www.edx.org/course/principles-machine-learning-microsoft-dat203-2x-6  ?

lucifer069:
Yes it is the one
https://www.edx.org/course/principles-machine-learning-microsoft-dat203-2x-6

eshellman:
Could you change the link, then?

lucifer069:
Changed the link

eshellman:
I think you've replaced the wrong link.

Issue #2494, opened 2017-07-10
Why does javascript-frameworks-resources.md mix resources?


eshellman:
hoping to get some historical perspective from @vhf 

Early on, a javascript frameworks resource list, including books, courses, tutorials, etc., was split off of the main free books list. At that time, there were no separate lists for courses, tutorials, etc.

https://github.com/EbookFoundation/free-programming-books/commit/7362e0671f29afe4991284d809c8fed30266d7af

One result is that contributors are confused about where js-related resources belong.

Given the very different complexion of the repo now, does it still make sense to split off js frameworks? 

Feedback from the community requested.

vhf:
I created this file because I was getting a lot PRs from people trying to add their favorite JS framework documentation or example blogpost to the `-books` files. It was an easy way of channeling all JS non-book content to a single place.

\~1 year later, @sindresorhus started the [awesome](https://github.com/sindresorhus/awesome/) project. Now that there are a bunch of good, well curated [javascript](https://github.com/sindresorhus/awesome/#programming-languages) lists, I'd say the need of maintaining a separate one in this repo doesn't make much sense anymore.

What I'd suggest is:

* Getting this repo better integrated in the _awesome_ lists ecosystem, occasionally redirecting there people trying to add to this project things that are not books,
* Removing the JS separate lists we have here in favour of the _awesome_ equivalents, potentially replacing the file content with a link as a manual, temporary redirection,
* More importantly, I wouldn't start adding things that are too far away from actual books to the books lists.

Which means that to the question

> does it still make sense to split off js frameworks?

I would say for: Yes, it makes sense for them to be separate things. It doesn't make much sense to maintain these files here though, as long as they are not books.

Up to you obviously. :)

PR #2710, opened 2018-02-05
Add Elm programming section

tpoindex:


eshellman:
Elm books, including IIRC one of these,  have been on the javascript-frameworks-resources page. We've decided it would be a good idea to re-merge books on that page with the free-books list: see #2688 and #2494. So if you could help this process along, the elm books would be easier to find.

tpoindex:
No problem, I'll check back after the re-merge of the javascript-frameworks-resources page.

eshellman:
Much appreciated!

PR #2711, opened 2018-02-05
add 3 books

brannx:


eshellman:
could you rebase from master so I can see what's new? also clojure with a j

PR #2713, opened 2018-02-06

brannx:


eshellman:
I think this url, for a site with 6 books, either belongs in the meta-list list, or the six books should be listed individually. In fact, one of the six books is already in the Bash list.

brannx:
Only few books in the section 'Linux' are about programming in linux, most of them are for sysadmins and this 6-books pack(yes, could be divided) comply with it. Maybe it is reasonable to embed all 'sysadmin' books in 'Language Agnostic' under new 'Linux System Administration' section.

eshellman:
Same is true of "Unix". I agree it makes sense to put sysadmin books in their own category.
Maybe:

* System Administration
  * Generic
  * Linux
  * Other
  * Unix

Also, think of how user will search - they won't think of Linux vs Linux Programming, they'll look for Linux vs Unix. And most books on sysadmin will have some programming, so drawing a distinction between linux programming and linux sysadmin could be confusing.

For the 6 linux-training books I prefer 6 listings, because people will inevitably want to add each separately.

brannx:
Also there is a book in this repo called 'Professional Assembly Language', I don't think it is free.

eshellman:
I agree.

Issue #2684, opened 2018-01-04
podcast-screencast file linting problem


eshellman:
the format being used for podcasts/screencast lists isn't understood by the linter - it only checks for sorting of the first string in brackets. We should change the format for these lists from

* [Podcast] - [Fragmented Podcast](http://fragmentedpodcast.com)
* [Screencast] - [Android Basics Archives - mybringback](http://www.mybringback.com/series/android-basics/)

to 

* [Android Basics Archives - mybringback](http://www.mybringback.com/series/android-basics/) (Podcast)
* [Fragmented Podcast](http://fragmentedpodcast.com) (Podcast)

brannx:
Ok

PR #2717, opened 2018-02-08
Update free-podcasts-screencasts-en.md

brannx:


eshellman:
Do you have an opinion on #2684 ?

PR #2722, opened 2018-02-09
Update urls of ‘Thinking in C++’

maluramichael:
The old links aren’t valid anymore


eshellman:
That's super that these are now on Archive, thanks! Worth looking to see if we have the other titles at https://archive.org/details/@bruceeckel

PR #2726, opened 2018-02-18
Added "Türkçe Git 101"

aliozgur:


eshellman:
Thanks!

eshellman:
And thanks for the book! Feel free to add the author name, it helps to distinguish books of similar name, and we want to recognize the authors.

PR #2491, opened 2017-07-08
Update free-courses-en.md

ShootingSpace:

PR #2732, opened 2018-02-23
add princeton algorithms

eshellman:
to close #2491


eshellman:
PR from @shootingspace needed alphabetization

PR #2734, opened 2018-02-25

brannx:


eshellman:
Nice addition! My only concern is that it seems as if it's not final. Would appreciate an update when it becomes final.

PR #2747, opened 2018-03-02

brannx:


eshellman:
thanks for doing this. I'll do the English list

brannx:
Consider this:
1. Separate podcats and screencasts and make two files free-podcasts-*.md and free-screencasts-*.md.
2. Stay free-podcasts-screencasts-*.md, but remove (Screencast) and (Podcast) hints.
3. Place 'Podcast' or 'Screencast' after each name. For example: RubyNoName Podcast.

eshellman:
1. maybe if the list got really long
2. the annotation can be optional
3. Don't like messing with the title.
(discussion on #2684)

PR #2750, opened 2018-03-03
Create javascript-frameworks-resources-ru.md

brannx:


eshellman:
see the discussion on #2494 
The conclusion is that we should merge resources on the javascript framework lists into the books lists and other lists as appropriate. Too confusing to maintain the separation.

PR #2755, opened 2018-03-04
Update README.md, good numbers

brannx:


eshellman:
The big milestone will the 1,000 contributors!

PR #2757, opened 2018-03-06
Add more information for BASH

sstt011:
This PR adds more information for BASH books where possible, including:
- Publication date for books more than five years old;
- Authors;
- Formats and alternative formats;
- and one URL fix.


sstt011:
Unfortunately I will not be able to respond over the next three days. I will reply as soon as I can.

eshellman:
Outstanding. #2383

PR #2758, opened 2018-03-07
Replace Dive into HTML5 link

rapfaria:
The old link (http://diveintohtml5.info/) returns Access Denied.


eshellman:
Great to see Creative Commons licenses at work.

PR #2769, opened 2018-03-15
Raspberry Pi and bit.ly

sstt011:
Add more information to Raspberry Pi books.
Also change bit.ly links to direct links.


eshellman:
I'm accepting this PR because these are good improvements. However, I think the "High Performance Computing" resource should be moved to the free courses list. (and not listed twice!) Thanks!

PR #2775, opened 2018-03-21
Move angular.js, node.js etc under javascript

sstt011:
Move various javascript apis/spinoffs/etc under javascript heading, like I did when I moved the javascript files from the javascript resources file.`


eshellman:
makes sense

PR #2778, opened 2018-03-28
Fixed broken Advanced Linux Programming link

wwaller:


wwaller:
Would it be better to instead link to the web archive link for the website, which still allows you to download the PDF? https://web.archive.org/web/20160324235624/http://advancedlinuxprogramming.com/

eshellman:
why not use https://mentorembedded.github.io/advancedlinuxprogramming/ ?

wwaller:
Oh, even better. Didn't see that.

PR #2779, opened 2018-03-29
add quality article

kompowiec:
lang & book choice, What to use when learning a programming language and how solve problems (very extensively and accurately)


eshellman:
Scanning the polish list, I see several items that aren't books, like this one (I'd put it in a "tutorial list"). It looks like a good one.

So

1. To make it easier if the list grows to the point where spitting the list into books, courses, tutorials, etc. could you add `(tutorial)` or the polish equivalent as an annotation?
2. lists need to be in alphabetical order.
3. Shouldn't we link to the original article, http://gynvael.coldwind.pl/?id=238 ?

kompowiec:
1. Sure.
2. Sorry i again forget about it, fixed
3. for me it makes no difference if mirror have original source

PR #2803, opened 2018-04-15
Fix smalltalk links

foochuanwei:
Fix broken links.


eshellman:
Thanks for this fix, and welcome to free-programming-books!

Issue #2785, opened 2018-04-04
"Opt-in required" in place of "Just fill the fields with any values"


packtpartner:
Can we add the notation "Opt-in required" instead of "Just fill the fields with any values" for email address required to get the free book wherever applicable?

eshellman:
Context for this: Several publishers of free ebooks want to collect email addresses of ebook users for marketing purposes. So in order to get a download link, users are asked to enter an email address. Publishers who do this include Packt, O'Reilly, Bookboon, Syncfusion, and Leanpub. Since before the Free Ebook Foundation was involved, Ebook listings for such items on Free-programming-books have been annotated  "*Just fill the fields with any values*", because while email addresses were requested, users declining to provide email addresses for whatever reason can just enter a fake email address and get the download link.

My view is that we need not discourage users from optionally joining a publisher's email list. By the same token, if an email address is really REQUIRED, the book is not really free; many users have valid reasons for not giving up a working email address, and reviewers shouldn't be forced to create an account to evaluate the worthiness of a submission.

I think the specific suggestion "*Opt-in required*" is problematic because if it's required, its not an "opt".

I think "*Email address requested*" would be acceptable. In any case, the notations should be applied uniformly, for fairness.

Other suggestions and opinions are welcome; we won't make a change unless there is rough consensus, or at least lack of major objection.

shunchu:
I agree with @eshellman's comment. If a user is required to give something in order to get another thing in return, then it's not exactly free. As such, it should be explicitly noted so that a user knows what s/he's getting into before clicking a link.

packtpartner:
Hi @eshellman,

So should we go ahead and make changes by adding "Email address requested" in our listings?
Also, to make it uniform do we have to make changes to all listings by changing "Just fill the fields with any values" to "Email address requested"?

eshellman:
The community should get at least a week to comment before a decision is made. #2776 should go with "just fill in the fields" for now; we'll make a blanket change in a new PR if that's what's decided.

packtpartner:
Okay, thanks!

eshellman:
implemented

packtpartner:
Hi eshellman,
Can we use "Email address requested" in our listings for future submissions?

PR #2811, opened 2018-04-24
Revise 'email requested' notes and policy

eshellman:
As discussed in #2785, we want to articulate a policy for books that ask for email addresses in order to enable downloads. While we want to encourage authors and publishers who wish to build communities around free ebooks, we recognize that some users may have important reasons to preserve privacy or may have limited email access.

This PR will stay open for at least a week to allow for community feedback and improvement.


eshellman:
The proposed changes:
1. change all `(just fill in fields with any values)` notations to `(email address *requested*, not required)`
2. add a note in CONTRIBUTING stating that we do not accept items which *require* working email addresses.

PR #2815, opened 2018-05-03
bad links removal needs verification 2018-05-02

borgified:
ive manually tested all the below removals but if someone would like to volunteer and double check that these are indeed broken links good to remove, please go ahead and double check.

i've also discovered that my script has some trouble with russian, japanese, chinese, and farsi languages. so i've avoided making any deletions for those md files until i get that resolved.

@eshellman

### DIFF
--- a/free-courses-en.md
+++ b/free-courses-en.md
@@ -32,7 +32,6 @@
 * [Misc](#misc)
 * [.NET](#net)
 * [OCaml](#ocaml)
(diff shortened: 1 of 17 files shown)


eshellman:
To avoid merge issues I'm going to accept this, then create an issue with a list of the deleted resources that people can help find and restore them where appropriate.

Issue #2358, opened 2017-05-18
create free-podcasts-screencasts-pl (polish) list 


eshellman:
With recent additions to free-programming-books-pl, there may be enough videos to start a free-podcasts-screencasts-pl list and move them. If any polish-speaking readers agree, it would be a great help to put together a PR.

charlottetan:
Suggestion: add Hacktoberfest label to this

PR #2822, opened 2018-05-15

danielkrupinski:
Removed dead links and added Python and Django books.


eshellman:
Thanks for these improvements, and welcome to free-programming-books. You're the 994th contributor to the project!

One of the items you've added is a course. Are there any other courses in the polish list that we would want to split off into a free-courses-pl list?

eshellman:
also see #2358

PR #2826, opened 2018-05-16
add some podcasts; also small reorganizing

brannx:


eshellman:
the mystery of cyrillic collation....

PR #2829, opened 2018-05-17

danielkrupinski:
Add C language book and course.


eshellman:
Great additions! Observation: Looks like the polish page does have many of the authors listed; we like to recognize them where appropriate.

PR #2830, opened 2018-05-18
Add Italian free courses page

mcicolella:


eshellman:
might be a bug in the linter

eshellman:
very strange linter behavior. But great to see italian add a courses page!

mcicolella:
Thanks. There is a lot of stuff to add!

PR #2856, opened 2018-05-24
add python tutorial and math

brannx:
Not sure where do "math for programmers" belong


eshellman:
it could also go in language agnostic:mathematics; Think Stats(for Python) is there.

PR #2860, opened 2018-05-25
Add Rails Testing Handbook

markoa:


eshellman:
Please see CONTRIBUTING.md: "We do not accept links to pages that *require* working email addresses to obtain books, but we welcome listings that request them.

PR #2863, opened 2018-05-26
Add Italian resources about 'Storia Informatica'

mcicolella:


eshellman:
These look like great resources, my main question is "Are they books?" 
I rely on Italian speaking contributors such as yourself to make these determinations somewhat consistently.

PR #2867, opened 2018-05-28
Update free programming books it

rozzilla:
I've added a PDF containing a Javascript course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,5 +87,5 @@ Free Podcasts and Screencasts:
 + [Free Programming Playgrounds](/free-programming-playgrounds.md)
 
 ## Licence
-Each file included in this repository is licenced under the Apache License 2.0.
-[Visit the official website of the licence](http://www.apache.org/licenses/) to know more about it.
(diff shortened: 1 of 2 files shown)


eshellman:
seems a bit old, don't you think?

rozzilla:
As I say in this cases: "Old but gold" 😀

eshellman:
What we've tried to do in other cases of "old but gold" is add a note of the year of publication. See #1926 If you can't find a year of publication, use an estimate. Thanks for the extra effort!

rozzilla:
Ok, I've added it ;)

README.md
```diff
@@ -85,3 +85,7 @@ Free Podcasts and Screencasts:
 
 ### Programming Playgrounds
 + [Free Programming Playgrounds](/free-programming-playgrounds.md)
+
+## Licence
```
eshellman: License for free-programming-books repo is CC BY

eshellman:
Thanks for this. Good idea to note the license in the README, but the British spelling is too confusing, given that github uses American spelling.

PR #2872, opened 2018-05-30
Remove broken links in Italian books

mcicolella:


eshellman:
The link for "Introduzione al Perl" might benefit from the adding the year (see discussion on #2867 )

mcicolella:
Done

PR #2876, opened 2018-06-03
add vga book also replacement

brannx:


eshellman:
category change seems reasonable; it's always helpful to explain the reasoning in case someone disagrees.

brannx:
At least linter is on my side :)

eshellman:
You have much linter-fu.

PR #2882, opened 2018-06-06
Remove R interactive tutorial

kakkar-sarthak:
Removed R interactive tutorial from the list as it redirects to [Pluralsight](https://www.pluralsight.com/search?q=R) instead of CodeSchool


eshellman:

PR #2888, opened 2018-06-09
add markdown resource

brannx:


eshellman:
That's not a book. https://spec.commonmark.org/current/ would qualify, though.

brannx:
But this site is dedicated to spec!?

eshellman:
the "books" pages don't list "websites" that aren't "books", because there are all sorts of websites. Some websites/webpages are like books - they have tables of contents, downloadable pdfs/epubs, etc.

PR #2894, opened 2018-06-14
add "Go 101" book

l2isbad:


eshellman:
Nice addition. The readme seems to imply that the book is unfinished, but it's not obvious if this is still true; if so, add an "in progress" note per CONTRIBUTING. Welcome to free-programming-books, we're very close to 1,000 contributors!

PR #2901, opened 2018-06-25
Adding Cormen's -Algorithms and Data Structures- 

AlexandraDrobut:
It is indeed a very good book, added it to the Language Agnostic section


eshellman:
While https://github.com/mitpress has a long list of free books https://mitpress.mit.edu/mit-press-open , this download does not appear to be authorized.

PR #2904, opened 2018-06-26

dzmitry-lahoda:
About author:
https://www.amazon.com/default/e/B001IXTSLU/ref=dp_byline_cont_book_1?redirectedFromKindleDbs=true
https://en.wikipedia.org/wiki/Mark_Burgess_(computer_scientist)

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -587,6 +587,7 @@
 
 #### Operating systems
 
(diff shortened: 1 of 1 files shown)


eshellman:
Nice addition. The linter wants to see a pdf indication. The author's site labels the book as "old"; consider adding a year indication.

PR #2914, opened 2018-06-29
Updated free-programming-books.md

kartikeytripathi:
Added Mastering Node.js in Node.js


eshellman:
Looks like this is a 7 year old unfinished book. I don't know enough about Node to know if it's still valuable What do you think?

PR #2950, opened 2018-08-26
add a ruby book

brannx:


eshellman:
I'm surprised we don't have this one - maybe it got removed previously. But definitely a good addition.

PR #2951, opened 2018-08-27
couple of old ruby books

brannx:
Maybe too old?


eshellman:
If you're not sure, add the year of last revision in the listing.

brannx:
They are both from 2009 and I think 10 years is too much, so I' m closing PR.

PR #2952, opened 2018-08-30
Add "Functional Programming, Simplified (Scala edition)" and "Hello, Scala"

alexanderfefelov:


eshellman:
I think it's a good idea to mark ~80% previews like that - we'll see if other folks agree.

PR #2961, opened 2018-09-15
Update Stanford iOS course to latest (iOS 11)

gusuy:
Replace an outdated course in the same Stanford series.


eshellman:

PR #2966, opened 2018-10-01
Added Arabic Java book

NurulKhairiah:
*Added "الشامل بغلة JAVA كامل


eshellman:
I'm uncomfortable with this link. The link prompted me to install flash, the download turned out to be a zip (which on some OS's can be a security risk) and the pdf it contained turned out to be a book from another web site with lots of added advertisements. On the other hand, the arabic list has no other books on java. Perhaps there's a cleaner link for this book, perhaps from the author's website?

PR #3027, opened 2018-10-19
Update link to a free version

Rastalamm:
Modifies a resource. 

Adds a link to the same book where you don't have to login/register at all.


eshellman:
Good change. consider adding the author's name.

PR #3033, opened 2018-10-22
Add author and remove duplicate resource

Rastalamm:
I've updated my previous PR to include the author (based on your suggestion)

I removed a duplicate entry for this book in the Django section. (Technically the book is about TDD across multiple languages but the Title states Python)


eshellman:
Thanks!

Issue #3040, opened 2018-10-24
add SICP russian translate (PDF)


fey:
Hello, propose to add russian translate of SICP 2nd edition
[Структура и интерпретация компьютерных программ ](http://newstar.rinet.ru/~goga/sicp/)
[pdf](http://newstar.rinet.ru/~goga/sicp/sicp.pdf) -  Гарольд Абельсон, Джералд Джей Сассман (PDF)
[PDF, optimized for screen reading ](PDF, optimized for screen reading )
to [free-programming-books-ru.md#0](https://github.com/EbookFoundation/free-programming-books/blob/master/free-programming-books-ru.md#0---language-agnostic)

eshellman:
1. should add authors in listing to better identify the book.
2. it that the translator's website? Is it linked from the SCIP page?

fey:
> 1. should add authors in listing to better identify the book.
> 2. it that the translator's website? Is it linked from the SCIP page?

1. I have updated OP 
Гарольд Абельсон -> Harold Abelson
 Джералд Джей Сассман ->Gerald Jay Sussman 

2. Yes it is translator's page, by [Георгий 'gogabr' Бронников](https://moikrug.ru/yura-bronnikov)
And source [another of russian translate](https://github.com/alexott/sicp-ru), but i dont know how it good.

PR #3042, opened 2018-10-24
add russian translate of SICP

fey:
add russian translate of SICP 
translator is Георгий Бронников


fey:
https://github.com/EbookFoundation/free-programming-books/issues/3040

PR #3054, opened 2018-11-06
Elm language resources added

alcole:
## What does this PR do?
Add Resource(s) 

### Description 
Elm language books, podcast and interactive editor added

Elm is fun!
most are official documentation, gitbooks or a wikibook
not so sure about this one: https://elmprogramming.com/ is on the open net but has a copyright declaration at the bottom of the page (non CC etc)

### Checklist:
- [ x] Not a duplicate
- [x ] Included author(s) if appropriate
- [ x] Lists are in alphabetical order


free-programming-books.md
```diff
@@ -39,6 +39,7 @@
 * [Elixir](#elixir)
   * [Ecto](#ecto)
   * [Phoenix](#phoenix)
+* [Elm](#elm)
```
eshellman: Good catch! Someone should check to see if we have any other missing headings.

eshellman:
Thanks for these additions. Some of these say that they're incomplete but seem mostly complete at quick glance. If you think they need an "in progress" notation (see CONTRIBUTING) you can add that as convenient.

PR #3305, opened 2019-06-28
Add Database Data Access for Highly-Scalable Solutions

dzmitry-lahoda:
## What does this PR do?
Add Resource(s)

### Checklist:
- [X] Not a duplicate
- [X] Included author(s) if appropriate
- [X] Lists are in alphabetical order
- [X] Needed indications added (PDF, access notes, under construction)

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -380,6 +380,7 @@
 #### Database
 
 * [Big Data Now: Current Perspectives from O'Reilly Radar](http://shop.oreilly.com/product/0636920022640.do)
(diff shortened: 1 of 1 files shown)


eshellman:
> We're sorry, this download is no longer available.

?

dzmitry-lahoda:
indeed - download fails. so html version is available. should i mark it as HTML? or better close pull? book is good, at leas for me.

eshellman:
I also get a "this content not maintained" banner. If you think it's worth adding, maybe an access note (html only) would help avoid confusion.

PR #3322, opened 2019-07-14
Remove Dead 'What You Need to Know about Docker'

dzmitry-lahoda:
## What does this PR do?
Remove Resource(s)

### Description 
`What You Need to Know about Docker - Scott Gallagher, Packt. ` is not found. No packt search helps.


MrS4w:
@eshellman

eshellman:
Thanks!

PR #3364, opened 2019-09-02
add 《Designing Data-Intensive Application》Chinese translate

cangfengzhs:
add 《Designing Data-Intensive Application》Chinese translate


eshellman:
This is an unofficial translation of a non-free book (there is an official translation, too: https://item.jd.com/12437624.html). If we are to list it, we should include the name of the original author, the name of the translator, and a label indicating it is an independent translation.

199911:
I guess this PR should be closed.
As it is neither a free book nor official translation.

PR #3377, opened 2019-09-23
replaced link from trial into pdf

dzmitry-lahoda:
## What does this PR do?
 Improve Repo

### Description 
Current 
image
and no download button.

i have replaced with link from author with download

so my link is free, previous is not

### Checklist:
- [x] Not a duplicate

It gives PDF  Version MOBI Version EPUB Version - so not sure what to do


eshellman:
we may need to check the other oreilly books

PR #3395, opened 2019-10-03
Revert 3021 master

jmanham:
remove


eshellman:
looks like you fixed one thing and added unneeded space on two things

PR #3413, opened 2019-10-15
Remove arch book because it's too old (about 2012) and link doesn't e…

numb95:
…xist

## What does this PR do?
Remove Resource  آرچ بوک in Persian

### Description 
The book link does not work (404) and also the book is about Archlinux book around 2012. so everything probably deprecated.


eshellman:
I think the index entry also needs removal

numb95:
@eshellman fix that

eshellman:
thanks

PR #3419, opened 2019-10-16
Added cloud computing interactive tutorials

tinabme:
## What does this PR do?
Add Resource(s)

### Why is this valuable
Cloud Computing a new and popular way of programming without the need for traditional devOps and server farms.

All of the beginning courses are free, and give users a great head-start on learning.

No these are interactive tutorials.

### Checklist:
- [√] Not a duplicate
- [√] Included author(s) if appropriate
- [√] Lists are in alphabetical order
- [√] Needed indications added (PDF, access notes, under construction)

### DIFF
--- a/free-programming-interactive-tutorials-en.md
+++ b/free-programming-interactive-tutorials-en.md
@@ -6,6 +6,7 @@
 * [C#](#c-sharp)
 * [C++](#c-1)
 * [Clojure](#clojure)
+* [Cloud Computing](#cloud computing)
 * [CoffeeScript](#coffeescript)
 * [Erlang](#erlang)
 * [Git](#git)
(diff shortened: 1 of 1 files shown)


eshellman:
These look really useful. However:
1. When I followed your EC2 link, I got non-free results. 
2. Would prefer links that went straight to individual courses. Otherwise maybe just have a single link to the site with the free filter turned on. 
3. since folks probably aren't familiar with quiklab, a notation '(registration required)' should be added.

eshellman:
Perhaps I don't understand something, but it looks like only the s3 course is free on quiklabs. The EC2 class is 1 credit, and the minimum spend is $8 for 8 credits.

tinabme:
> Perhaps I don't understand something, but it looks like only the s3 course is free on quiklabs. The EC2 class is 1 credit, and the minimum spend is $8 for 8 credits.

You are correct, I misunderstood the intro courses to be free, sorry. Now all of the courses listed are free.

eshellman:
Thanks!

PR #3420, opened 2019-10-18
Updated "Dive Into Python" links

zackdever:
## What does this PR do?
Update Resource

## For resources
Mark Pilgrim's Dive Into Python (both 2 and 3)

### Description 
The "official" Dive Into Python links no longer go to the books. There is no great authoritative link to use since Mark Pilgrim left the internet (https://en.wikipedia.org/wiki/Mark_Pilgrim#%22Disappearance%22_from_the_Internet), but these are returned high in Google results and appear to load really quickly.

The Dive Into Python books are still valuable resources, and there are plenty of good mirrors still available. As it stands right now, the existing links go to pages that have nothing to do with Dive Into Python.

The book is just a pretty plain HTML site you can view at the updated links.

Yep.

### Checklist:
- [ x] Not a duplicate
- [ x] Included author(s) if appropriate
- [ x] Lists are in alphabetical order
- [ x] Needed indications added (PDF, access notes, under construction)


eshellman:
Nice update - these are valuable resources. Welcome to free-programming-books!

Issue #3357, opened 2019-08-21
Broken Spanish book link


aalonso99:
Links to Spanish "Aprenda C++ avanzado como si estuviera en primero" and "Aprenda C++ básico como si estuviera en primero" are broken. 
Also I checked the documents in other websites, none of them official, and it is from Navarra's University. I cannot find the originals and there is no note about their license, so I am not sure if they really are open and free to use. 

Thank you for maintaining this repo :)

rosepac:
> Links to Spanish "Aprenda C++ avanzado como si estuviera en primero" and "Aprenda C++ básico como si estuviera en primero" are broken.
> Also I checked the documents in other websites, none of them official, and it is from Navarra's University. I cannot find the originals and there is no note about their license, so I am not sure if they really are open and free to use.
> 
> Thank you for maintaining this repo :)

I don't understand what the real problem is. Nor what you do not understand. 

The freedom of use of the file in question is beyond doubt, since the file is written in the form of a university article; and the author himself exposed it to the public without any problem, for its use.

The fact that the author wanted to take that university article and transform it into an edited book.

That would not change the original file, nor its right to use. It is not a public domain book, nor under a creative commons license. In fact, it is not a book.

It is a university article. So it can be used for personal, learning or study use of the subject in question. Without any kind of problem.

In fact, the author has many other articles of similar nature, and in no case did he expose any type of license within them.

That same act. It defines us, because and for what, those files have been created? 

The only thing I miss in this case. Is that within the author's list of articles. This one in particular is marked as "Ruano Training". It is supposed to be a sutdy academy. What is absolutely impossible. Because you are not going to create a book for an academy with all the official typing of the university.

In any case, they are books that are surely out of date.  And if the author would like to license them. They would be new versions of the documents of that time and under another type of edition. I understand In this case there is no book or edited, nor of the author, nor in the redirection of the page of the University..

If we look a little deeper. The author, happened to the university [of madrid in 2002](https://cima.unav.edu/detalle-noticias-inmunologia-e-inmunoterapia/-/asset_publisher/8Ha8/content/2019_06_03_tecnun-javier-garcia-de-jalon-ingeniero-laureado-por-la-real-academia-de-ingenieria/10174?tituloNoticia=javier-garcia-de-jalon-ingeniero-laureado-por-la-real-academia-de-ingenieria&fechaNoticia=03-06-2019)

The document is used by other universities. 
[...]

PR #3422, opened 2019-10-19
update links for #3357

pritam-patil:
- The links were found on google by doing keyword
  search using filetype and inurl
- The links and files verified

## What does this PR do?
Update Resource(s)

- Broken links are fixed.

- Files are downloaded and verified for content.

### Checklist:
- [x ] Not a duplicate
- [ ] Needed indications added (PDF, access notes, under construction)

### DIFF
--- a/free-programming-books-es.md
+++ b/free-programming-books-es.md
@@ -136,8 +136,8 @@
 
 ### C++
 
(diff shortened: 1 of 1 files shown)


eshellman:
could you check if these are in archive.org wayback machine? because it looks like both these links point at university courselists, and courses tend to change from year to year. Thanks!

Issue #2990, opened 2018-10-09
(Java section) Object-Oriented vs. Functional Programming don't work


harrykar:
For your information 
In Java section 
[Object-Oriented vs. Functional Programming](http://www.oreilly.com/programming/free/object-oriented-vs-functional-programming.cspl)'s  distribution seems not work i haven't yet received any mail with download link as they(oreilly) say

PS : 
You refer: _(email address requested, not required)_ That's not correct i had to provide my email address bot being able to press the download button and they 
"respond"  
Thank you.
Enjoy this free ebook from O’Reilly Media. Check your email for a link to download.

harrykar:
I have found it here 
http://www.r-5.org/files/books/computers/languages/java/scala/Richard_Warburton-Object-Oriented_vs_Functional_Programming-EN.pdf

eshellman:
If O'Reilly books require a working email, we may need to remove the listings.

ShipsWithCannons:
The page currently 404's. So in this case, the link will have to be removed regardless.

bluetata:
@harrykar posted a wrong link for the book, In our repo the book link is http://www.oreilly.com/programming/free/object-oriented-vs-functional-programming.csp   

**For** the book link in our repo. I provided my email address, then   

![image](https://user-images.githubusercontent.com/25427352/50872378-6b08c780-13fa-11e9-997c-ce2fa6388600.png)
![image](https://user-images.githubusercontent.com/25427352/50872384-70fea880-13fa-11e9-89b1-8f13f7955a88.png)

**BUT** I haven't also received any mail with download link - -! Does anyone else have the same problem as me?

pritam-patil:
Hi guys, I've updated the .csp link from oreilly.com to a .pdf link where it's free and contains right content. Do review [this PR](https://github.com/EbookFoundation/free-programming-books/pull/3423).

And thank you for creating and maintaining this repo!

PR #3423, opened 2019-10-19
Update a Java book URL - #2990

pritam-patil:
Changes:
- The book 'Object-Oriented vs. Functional Programming' has a .csp link
- Also a sign-up is required to read the book since it's hosted at oreilly.com
- The contributing guide is against a hosting service and required sign-up.
- That is why this PR is created.

## What does this PR do?
Update Resource(s)

The .csp link is replaced with .pdf link.
Also sign-up is avoided to read .csp page.

The pdf is downloaded free and verified for correct content.

Yes

### Checklist:
- [ ] Not a duplicate


eshellman:
We've had a bunch of O'Reilly titles go dark, but "pepe.holla.cz" doesn't appear to be an authorized distribution for this book, and if it's asking for user emails, that especially suspicious. We can't include this link without better documentation - better to delete the O'reilly link.

PR #3459, opened 2019-12-01
Remove "Wizard code" 

rndd-AK:
## What does this PR do?
Remove "Wizard code" from asm section

Because there is no book.


eshellman:

PR #3460, opened 2019-12-01
Backbone js was duplicated

Uzay-G:
Removed it from general js section so it is now only in backbonejs section


eshellman:

PR #3461, opened 2019-12-02
Added StackBlitz under Angular in free-programming-playgrounds page

AnkitSharma-007:
## What does this PR do?
Add Resource(s) 

StackBlitz is a free online code editor for Angular. This is not included in the existing link.

### DIFF
--- a/free-programming-playgrounds.md
+++ b/free-programming-playgrounds.md
@@ -28,6 +28,8 @@
 ### Angular
 
 * [Plunker](http://plnkr.co)
+* [StackBlitz](https://stackblitz.com/)
+
 
 
 ### ClojureScript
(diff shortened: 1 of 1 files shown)


eshellman:
I'm not sure how we should deal with the other app flavors supported by StackBlitz. I guess we'll deal with it later. Welcome to free-programming-books!

eshellman:
whoops - failed to notice the broken build

AnkitSharma-007:
Thanks for accepting the PR. Maybe we can add StackBlitz under other frameworks also.

eshellman:
we try to avoid having duplicates. If there are different links for each framework, that avoids the problem.

AnkitSharma-007:
Yes, we have separate URL for StackBlitz playground of Angular, React, Ionic, Typescript, RxJs and Svelte. Shall i raise another PR to include these links?

eshellman:
yes please!

PR #3463, opened 2019-12-03
Added StackBlitz playground URL

AnkitSharma-007:
## What does this PR do?
Add Resource(s) 

### Description 

Added StackBlitz playground URL for Angular, Ionic, React, Svelte and Typescript.


eshellman:
I think we may want to organize the javascript frameworks into a category with subcategories to mirror the book list, but we'll leave that for another PR.

PR #3483, opened 2019-12-19
Added an Archive.org link for a missing book

HankiDesign:
## What does this PR do?
Improve Repo

### Description 
The previous link for _Data Structures and Algorithms: Annotated Reference with Examples_ doesn't work anymore so I updated it to point to the original page on archive.org. The book PDF has also been scraped by archive.org so the download link works.

I was able to find another direct link to the PDF on the same domain as the previous link (mdp.ac.id) but that seems to be a very random source (Indonesian school or education portal, can't tell for sure) not related to the original book at all. Since one of the first pages in the book mentions that it's exclusively published on dotnetslackers.com, I think it's fair game to keep the original site as the download page.

### Checklist:
- [ ] Not a duplicate
- [ ] Needed indications added (PDF, access notes, under construction)


eshellman:
The source explanation is appreciated.

PR #3506, opened 2020-01-09
Added Flavio Copes books

watkinsmatthewp:
## What does this PR do?

Add Resources

### Description 

Flavio Copes's ebooks on the list now

More free ebooks :)

I just downloaded them. For free. From the author.

Yes

### Checklist:
- [ ✔] Not a duplicate
- [ ✔]  Included author(s) if appropriate
- [ ✔] Lists are in alphabetical order
- [ ✔] Needed indications added (PDF, access notes, under construction)


eshellman:
So we need to agree to get a weekly newsletter to get the "free" ebooks? And the site has no privacy policy? Are the books THAT good?

watkinsmatthewp:
I've only been reading a few of them but I like the books. I started following his articles, too.

You don't _technically_ need to sign up to get his books. He hosts them as static files on his site accessible to anyone, but the URLs are not published anywhere on his site-- they are sent in an email. I could have pasted direct links to the ebook files and marked the books as email _requested_ since that's technically true, but that's not really true to the intent of the author.

eshellman:
Right, we don't want to use direct urls if they're not on the website. If the author doesn't want it to be free, it isn't free.

My preference is to discourage the model where you have to sign up for an email list to get an ebook. If you've been on the list long enough to know that you're not going to be constantly asked to buy something and can remove the email from the list for reals - as would be required by GDPR and the California privacy laws, then we can add the books. Since the author is apparently writing a "how to have a successful blog" book, we don't want to give the wrong idea!

A real issue for this list is knowing whether the book still exists. If the author loses interest after a year, there's no easy way to check if the ebook sender is dead. Yes, authors tend to lose interest and leave ghost websites,

I also wish that the landing pages gave more context for folks to know if they want the books in exchange for the email address.

I'll leave the PR open for a while to see if other F-P-B contributors have any comments. Feel free to poke the PR once I forget about it. Thank you for making the effort to add the books and satisfy the linter, the effort is appreciated.

watkinsmatthewp:
If the email requirement is really a problem, then sadly, a lot of the books already on the list would not make the cut. For example, I actually just downloaded "TypeScript for C# Programmers" because my company is looking to use TypeScript for our next project and the author (InfoHQ in this case) required me to not only subscribe temporarily to a mailing list but to create an account.

What are the requirements for getting books on the list? Is is only supposed to be email-requested or lower? I'm a little confused...

Thanks!

watkinsmatthewp:
Updated the PR merge to include changes from master

199911:
I think we should close this CR

PR #3517, opened 2020-01-16
Update free-programming-books-ar.md

ItIsMeCall911:
## What does this PR do?
Add Resources

Yes
### Checklist:
- [x] Not a duplicate

### DIFF
--- a/free-programming-books-ar.md
+++ b/free-programming-books-ar.md
@@ -62,8 +62,8 @@
 
 ### Python
 
-* [إحترف بايثون الآن](https://docs.google.com/viewerng/viewer?url=https://books-library.online/files/download-pdf-ebooks.org-1521324052-896.pdf) (PDF)
 * [أتعلم بايثون](https://docs.google.com/viewerng/viewer?url=https://books-library.online/files/books-library.online_noo72561738871c402b720af1-6336.pdf) - أرض وينو (PDF)
+* [احترف بايثون الآن](https://docs.google.com/viewerng/viewer?url=https://books-library.online/files/download-pdf-ebooks.org-1521324052-896.pdf) (PDF)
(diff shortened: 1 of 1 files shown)


eshellman:
some nice resources! However...
1. please don't use the viewer links, use direct links. people will have their own preferences for pdf viewing.
2. "download2.org" appears to be a file hosting site. We accept this sort of link when there is evidence that the book's author/publisher has authorized the posting. 
3. arabic titles are "alphabetized" using the unicode sort order provided by python.

Thanks for persevering!

eshellman:
From https://github.com/EbookFoundation/free-programming-books/blob/master/CONTRIBUTING.md  - "we don't accept files hosted on google drive, dropbox, mega, scribd, issuu and other similar file upload platforms"

eshellman:
please remove the hassouna-academy.com and noor-book.com books and the PR is fine. Those two sites don't even provide the books they list.

ItIsMeCall911:
> please remove the hassouna-academy.com

Didn't u ask for the original source ? hassouna-academy.com is the original source for "صخرة الأساس ملخص لغة Python " and the book is available to download from there.
As for noor-book.com all the books i listed are also available to read/download.
Let me end my comment with a little bit of clarification, every book i listed are 100% free and the reason those books are on sites like noor-book.com instead of the original source it's because most of sites/blogs of the original books are dead by now.
Hope this make things clear.

eshellman:
The link to hassouna academy: https://www.hassouna-academy.com/ lists the python book with a link to http://www.mediafire.com/file/cpzkhp5xnbzv42v/basic-rock-python-short.pdf/file
mediafire links to http://download1647.mediafire.com/902857yaxqcg/cpzkhp5xnbzv42v/basic-rock-python-short.pdf Mediafire is a file hosting site, however, since the original source of the book is linking to it, the hassouna academy link is acceptable.

The Noor-book links are not acceptable. There is no evidence that the authors have given permission to Noor-book to distribute the books. A look at the Noor-book home page shows many best-seller books that are obviously not authorized for free distribution. If these books come from dead sites, check internet archive wayback machine for availability.

Thanks for bearing with us on issues like these - we need to be careful about unauthorized distribution to avoid legal problems.

ItIsMeCall911:
> we need to be careful about unauthorized distribution to avoid legal problems.

That's why i linked pdf first instead of Noor-book because i know that site also contain unauthorized book distribution
And as for wayback machine that's a dead end for me cuz it will be a waste of time/effort searching for the original url (from nowhere) and then check if it was archived properly when i already know that any search will lead either to a direct pdf link or to a book piracy site like Noor-book, which is explain why this Index has not been updated for more then year ☹
Also a thing worth noting which is "author uploading his book to a piracy/file sharing site" is a common thing here because people also view those sites as a way to host their books.
i already removed all Noor-book links so am gonna close this pull request

PR #3526, opened 2020-01-19
Update free-programming-playgrounds.md

natkuhn:
Added codesandbox.io under React (and removed the react reference from its listing on JavaScript).
Added a Java section and linked to repl.it
repl.it actually has sandboxes for most of the languages listed here, and maybe some others, but I didn't add them all in. The Swing support for Java is new, and has some bugs. They didn't respond to my report, so I think they are spread a little thin.

## What does this PR do?
Add Resource(s) [repl.it], Improve Repo [copied codesandbox.io to #react]

I have used both codesandbox and repl.it

### Checklist:
- [x ] Not a duplicate
- [x ] Included author(s) if appropriate
- [x ] Lists are in alphabetical order

### DIFF
--- a/free-programming-playgrounds.md
+++ b/free-programming-playgrounds.md
@@ -141,6 +141,7 @@
 * [CodeSandbox.io](https://codesandbox.io)
 * [StackBlitz](https://stackblitz.com/fork/react)
 
+
 ### Ruby
 
 * [Codepad](http://codepad.org)
--- a/free-programming-playgrounds.md
+++ b/free-programming-playgrounds.md
@@ -9,6 +9,7 @@
 * [Go](#go)
 * [Haskell](#haskell)
 * [Ionic](#ionic)
+* [Java](#java)
 * [JavaScript](#javascript)
 * [Kotlin](#kotlin)
 * [.Net](#dotnet)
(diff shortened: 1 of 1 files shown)


eshellman:
Do you think we should group the react and Angular links as subcategories of Javascript to mirror our books list? That way we could remove the duplicate listing.

natkuhn:
@eshellman That makes a lot of sense. I just happened on your site and so I will not be creating PRs for everything I see, but honestly I found it quite confusing. So here is some feedback, to use or ignore as you see fit.

1. It is not clear whether this is for "books" or "resources". I vote for making it "resources." Then the sections would be Books, Online Courses, Interactive Programming Resources, Problem Sets, Podcast - Screencast, Programming Playgrounds (though could that be combined with "Interactive Programming Resources"?)

1. At the very least, "View the English list" should say "View the English list of books".

1. The pages for "Online Courses", "Interactive Programming Resources", "Programming Playgrounds" are just labeled "Index" so it is impossible to know what you are looking at. They should be titled "Index of Programming Playgrounds", e.g.

1.  I agree that it would be good to organize all the pages the same way, and it may make sense to list frameworks under languages, but then it would be good to indent the frameworks on the full list, rather than have them as [top-level entries the way they are here](https://ebookfoundation.github.io/free-programming-books/free-courses-en.html#angularjs). That may not be easy to do in Jekyll. I see that the titles are `<h3>`s. Maybe the subsection titles could be `<h4>`s with some css to indent them. It would be OK to just indent the subheader.

1. Since I first posted this, I used repl.it as an online Python playground. It works great.

Thanks for keeping up such a great resource, sorry I don't have time to do the PRs myself!

eshellman:
Thanks for the suggestions- it would be great if you moved them to issues, If sure someone will make PRs if they agree with your points. You're right, "books" has grown into "resources", but even with a thousand-plus collaborators, there are too many "resources" to make a useful list.

natkuhn:
Thanks, I just did that. I didn't include adding repl.it to the Python playgrounds, since that seems to fall here. Best, Nat

nurik1999:
can I add css file no java swing

nurik1999:
please. help me

eshellman:
seems I forgot to merge this. not sure what @nurik1999 wants. Welcome @natkuhn !

PR #3530, opened 2020-01-20
Fixed kotlin id link

watkinsmatthewp:
This commits fixes the casing of the header ID for the Kotlin link (apparently the only language where it was incorrectly cased).


eshellman:
thanks1

PR #3582, opened 2020-04-15
add angstromctf.com

ukH9YFxczKfxEnNECPyq:
## What does this PR do?
Add Resource

### Why is this valuable
Just another set of cyber security problems


eshellman:
Thanks!

PR #3662, opened 2020-08-01
Add awesome bot

borgified:
## What does this PR do?
- added automation to check urls: to trigger check, add `check_urls=free-programming-books.md` to your commit message, then look at the details of your travis checks.
- existing fpb-lint will continue to work as normal
- takes about 5mins to run
- example run: https://travis-ci.org/github/borgified/free-programming-books/jobs/715754535
- you can specify what file(s) to check, just put it after the `check_urls=` as a space separated list of files (eg. `check_urls=free-programming-books.md free-programming-books-en.md`)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ jobs:
   include:
     -
       language: node_js
-      sudo: false
       node_js:
         - 6
       before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,4 +20,6 @@ jobs:
       before_script:
         - gem install awesome_bot
       script:
-        - awesome_bot $(echo $TRAVIS_COMMIT_MESSAGE | awk -F: '{print $2}') --allow-redirect --allow-dupe --skip-save-results
+        - FILENAME=$(echo $TRAVIS_COMMIT_MESSAGE | awk -F: '{print $2}')
+        - echo $FILENAME
+        - awesome_bot "$FILENAME" --allow-redirect --allow-dupe --skip-save-results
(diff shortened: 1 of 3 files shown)


eshellman:
Interesting! I'm assuming this is not intended for merging right now - if nothing else it would need additions to documentation. Would be really nice if a pr could specify a file to check.

borgified:
ok ill try to make it so that the file to check can be specified via commit message

borgified:
all set @eshellman. the commit message to trigger changed a little though. this is so that you can specify more than one file to check.
for example, if your commit message is:
```
check_urls=free-programming-books.md free-programming-books-en.md
```
then this would make it check both free-programming-books.md  AND free-programming-books-en.md

one thing to note is that the build result is now based on the last checked file. so if you look at the current checks, you'll see that it should've actually failed cuz there are link issues with the first file (free-programming-books.md) but because the last file had no problems, then it makes the build green.

basically you just put "check_urls=" first then a space separated list of files to check and awesome_bot will try to check the urls within the specified files. (hope it works for languages other than english, it probably should)

borgified:
also not sure where you want this documented. it'd just be pretty much copy paste of what ive written above.

eshellman:
Add a section in CONTRIBUTING, I guess. copy paste looks like most of it.

jc-mindstronghealth:
ah i forgot about that file. good call. i've added a short blurb, let me know if it's clear enough.

borgified:
ah i forgot about that file. good call. ive added a short blurb, let me know if it isnt clear enough

eshellman:
Thanks! Will be really interesting to see how this works.

PR #3680, opened 2020-08-24
fix broken urls

borgified:
## What does this PR do?
Remove Resource(s)

also

it shows the new check_urls in action. @eshellman see "checks" section below, click Details, and inspect logs to see broken urls listed.

It's still failing because there are some other links that are still causing problems due to a variety of issues like:
- took too long to load (timeout)
- weird server response
- other false positives?

but at least we cleaned up a bunch of real broken links so we can just hit merge despite the check_url failure.

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -437,6 +437,7 @@
 * [Introduction to Data Science](https://docs.google.com/file/d/0B6iefdnF22XQeVZDSkxjZ0Z5VUE/edit?pli=1) - Jeffrey Stanton
 * [Introduction to Data Science](https://leanpub.com/datasciencebook) - Rafael A Irizarry *(Leanpub account or valid email requested)*
 * [Mining of Massive Datasets](http://www.mmds.org)
+* [School of Data Handbook](http://schoolofdata.org/handbook/)
 * [Statistical inference for data science](https://leanpub.com/LittleInferenceBook/read) - Brian Caffo
 * [Theory and Applications for Advanced Text Mining](http://www.intechopen.com/books/theory-and-applications-for-advanced-text-mining)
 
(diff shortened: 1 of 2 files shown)


free-programming-books.md
```diff
@@ -439,7 +437,6 @@
 * [Introduction to Data Science](https://docs.google.com/file/d/0B6iefdnF22XQeVZDSkxjZ0Z5VUE/edit?pli=1) - Jeffrey Stanton
 * [Introduction to Data Science](https://leanpub.com/datasciencebook) - Rafael A Irizarry *(Leanpub account or valid email requested)*
 * [Mining of Massive Datasets](http://www.mmds.org)
-* [School of Data Handbook](http://schoolofdata.org/handbook/)
```
eshellman, line 442: good url but uses cloudflare robot blocking.

free-programming-books.md
```diff
@@ -1947,7 +1938,7 @@ Kerridge (PDF) (email address *requested*, not required)
 
 * [HGInit - Mercurial Tutorial](https://web.archive.org/web/20180926172759/http://hginit.com/) - Joel Spolsky
 * [Mercurial: The Definitive Guide](http://hgbook.red-bean.com)
-* [Mercurial: The Definitive Guide 2nd edition](https://book.mercurial-scm.org) ([source repository](https://bitbucket.org/hgbook/hgbook)) (:construction: *in process*)
```
eshellman, line 1950: LGTM

eshellman: never mind - it's the second url

free-programming-books.md
```diff
@@ -2261,7 +2251,6 @@ Kerridge (PDF) (email address *requested*, not required)
 * [Introduction to Programming with Python](http://opentechschool.github.io/python-beginners/en/) (3.3)
   * [Introduction to Programming Using Python](http://python-ebook.blogspot.co.uk) - Cody Jackson (1st edition) (2.3)
 * [Introduction to Python](http://kracekumar.com/post/71171551647/introduction-to-python) - Kracekumar (2.7.3)
-* [Invent Your Own Computer Games With Python](http://inventwithpython.com/chapters/) - Al Sweigart (3.1)
```
eshellman, line 2264: http://inventwithpython.com/invent4thed/

eshellman:
not bad. only one false negative. wondering if we need a (no robots) annotation. Is this supposed to still return an error?

free-programming-books.md
borgified: wow.. thats annoying.. didnt realize i had to manually click on that captcha

borgified:
when it hits that cloudflare robot blocking, it returns a 503 (Service Temporarily Unavailable) which makes it look like the website is down. unfortunately that makes distinguishing websites that are actually down.

there **is** a whitelist feature with the check_urls but keeping track of that would become a maintenance burden

eshellman:
Is there any possibility we could have the checker commit a list of failing urls instead of returning an error? @borgified If it's OK with you, I'll add you to the maintainers group so you can experiment.

PR #3697, opened 2020-09-02
Move a playground link to CSS section

jgpacker:
FlexBox is a CSS feature.
This move is consistent with other files.

## What does this PR do?
Improve Repo

### DIFF
--- a/free-programming-playgrounds.md
+++ b/free-programming-playgrounds.md
@@ -20,6 +20,7 @@
 * [R](#r)
 * [React](#react)
 * [Ruby](#ruby)
+* [RxJS](#rxjs)
 * [Rust](#rust)
 * [Scala](#scala)
 * [SQL](#sql)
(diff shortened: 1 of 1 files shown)


eshellman:
I agree with moving the link. linter wants x to come after u. The broken link is not yours- fixing it is optional.

jgpacker:
@borgified from my perspective:
* It can be confusing to have checks from Travis CI and Github Actions at the same time, because each check has **logs** in a different place (by the way, thanks for the comment @eshellman ).
* As far as I understood, `check_urls` checks URLs only from files changed in the pull request. I think that's neat! To make this clearer, I recommend changing it's name from `check_urls` to something like `Check URLs from changed files`.

eshellman:
Thank you @jgpacker, both for your PR and your perspective on check_urls.

@borgified I think that running check_urls is really useful, but returning a "fail" is not helpful. added urls get checked for content anyway. can the results of check_urls could be recorded in a separate file somehow?

PR #3702, opened 2020-09-05
Improve url checker

borgified:
- make the check more descriptive
- avoid failing build even if broken urls are found
- save results into a file
Screen Shot 2020-09-04 at 11 28 15 PM


borgified:
i cant do much about the fact that the logs for travisci vs github workflows are accessed in a different manner, they are two different systems and have inherent differences. i could've implemented these changes in travisci but i wanted to learn the github workflow way of doing things which i think has some advantages since it is better integrated to github.

eshellman:
looking forward to seeing how this works!

PR #3789, opened 2020-10-07
im added some resource 

dfr-hub:


eshellman:
why do you think this is a "book"?

dfr-hub:
hello @eshellman 

 I've edited the title for this book
why I named this book because I feel it is a book

eshellman:
1. we prefer atomic PRs.
2. see travis-CI output for formatting alerts

PR #3803, opened 2020-10-08
Adds Algebraic Topology by Allen Hatcher

kagof:
## What does this PR do?
Add Resource(s)

### Description 

This adds the book Algebraic Topology by Allen Hatcher. <https://pi.math.cornell.edu/~hatcher/AT/ATpage.html>. Note that I linked to the home page for the book; this lets people access the errata, get the print version (which is not free, but is cheap), or download individual chapters. We could instead directly [link to the PDF](https://pi.math.cornell.edu/~hatcher/AT/AT.pdf), if that is preferred.

This is a great text that was used for my Algebraic Topology course in uni. I don't see any other books on this topic in the list, so, although it is rather abstract, it may still be useful. Also, the author is diligent about making corrections even to this day.

The page I linked to contains the quote from the author:

> By special arrangement with the publisher, an online version will continue to be available for free download here, subject to the terms in the copyright notice.

[The copyright notice](https://pi.math.cornell.edu/~hatcher/AT/ATcopyright.html) simply says

> Single paper or electronic copies for noncommercial personal use may be made without explicit permission from the author or publisher. All other rights reserved.

### Checklist:
- [x] Not a duplicate


eshellman:
Thanks for the excellent PR and welcome to free-programming-books!

PR #3859, opened 2020-10-09

AnkitSharma-007:
Created a new category for ASP .NET Core and added a new book to it.

### DIFF
--- a/free-programming-books.md
+++ b/free-programming-books.md
@@ -8,6 +8,7 @@
 * [APL](#apl)
 * [Arduino](#arduino)
 * [ASP.NET](#aspnet)
+* [ASP.NET Core](#aspnet-core)
 * [Assembly Language](#assembly-language)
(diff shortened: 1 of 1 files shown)


eshellman:
not sure its helpful for users to separate the category. and that link is not a good one to use

AnkitSharma-007:
ASP .NET and ASP .NET Core are two different frameworks. Therefore, I have separated the categories. 

Regarding the link, The author has hosted the book on GitHub so that he can update the contents with time. I am afraid this is the only link currently available for the book. 

Do you have any suggestions regarding this?

eshellman:
one is described as a version of the other. not helpful to separate them unless the list is longer. 

ask the author to provide a better link

AnkitSharma-007:
Unfortunately the author cannot provide any alternative links. I am closing the PR. 

Thanks.

PR #3890, opened 2020-10-10
Added more course links for databases

Swapnil123123:
## What does this PR do?
Add Resource(s) 

### Description 
Provided links to good free courses for (mySql) and (mongoDB). One can add many more databases in this category.

Knowledge on databases is very essential since programming is not just about frontend. Rather, it is a combination of frontend, middleware and backend to make a functional application. 

These courses are free and I have carefully selected these courses. One can even visit the links to find out.

No

### Checklist:
- [x] Not a duplicate

### DIFF
--- a/free-courses-en.md
+++ b/free-courses-en.md
@@ -166,7 +166,7 @@
 #### mongoDB
 
 * [Introduction to MongoDB](https://www.tutorialspoint.com/mongodb/index.htm) 
-* [MongoDB University](https://university.mongodb.com/)
+* [MongoDB University](https://university.mongodb.com)
 
 
 #### mySql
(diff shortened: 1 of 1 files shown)


eshellman:
There's also an SQL section. Maybe that belongs too? also check the organization of the books list, as having one system makes it easier for users

Swapnil123123:
> There's also an SQL section. Maybe that belongs too? also check the organization of the books list, as having one system makes it easier for users

I see Sql section is present, but shouldn't there be a common section called database under which subcategories should contain names of databases & relevant links?

Swapnil123123:
I can remove the sql section n place it under databases section instead

eshellman:
Think about how people would look for these things. having all the db books together makes sense.

PR #4066, opened 2020-10-14
Update free-courses-en.md

scope2018:
Added a free computer systems security course from MIT OCW


eshellman:
Please add instructor names where appropriate.

PR #4543, opened 2020-10-20
Added formats to resources

giuseppebrb:


eshellman:
we don't really need (HTML) indicators, except when a resource is ONLY HTML. But it seems you haven't checked. the TeX resource is available in pdf, postscript, and TeX.

PR #4561, opened 2020-10-21
ADD: Redis PPT by Eko

shidqi:


eshellman:
this appears not to be a book. 
In general, we do not link to file sharing sites.

Issue #4892, opened 2020-10-30
Found some duplicate and redirecting Links


U-C-S:
Example in Bash and C#:
<pre>
* [Bash Notes](https://goalkicker.com/BashBook) (PDF)
* [Bash Notes for Professionals](http://books.goalkicker.com/BashBook/) - Compiled from StackOverflow documentation (3.x) (PDF)
</pre>
<pre>
* [C# Notes](https://goalkicker.com/CSharpBook) (PDF)
* [C# Notes for Professionals](http://books.goalkicker.com/CSharpBook/) - Compiled from StackOverflow documentation (3.x)
</pre>

Here, books.goalkicker.com is re-directed to goalkicker.com

And

This books.goalkicker.com is used in few other places
<pre>
* [Git Notes](http://books.goalkicker.com/GitBook) (PDF)
</pre>
I suggest to go with the original: Goalkicker.com for all the redirects.

eshellman:
yes please!

PR #4898, opened 2020-10-30
Removed Duplictes/redirect links to goalkicker.com & Added few books from same

U-C-S:
## What does this PR do?
- Removed redirecting links for goalkicker.com (#4892)
- Added links for some programming languages from goalkicker.com

### Checklist:


eshellman:
Make sure all the links have the notation `- Compiled from StackOverflow documentation`
Probably we should remove the version note. `(3.x)`
IIRC the goalkicker books are also available as EPUB, I suggest just removing the `(PDF)` notations if  that's the case.
Thanks so much for taking this on.

U-C-S:
@eshellman Now, Review it.

I found that goalkicker books aren't available as EPUB format, So, I left PDF tag as it is.

eshellman:

PR #5057, opened 2020-11-03
sorted the vietnamese lists

eshellman:
for some reason the linter doesn't check the vietnamese file


eshellman:
Would appreciate comment from @LeoTino

PR #5142, opened 2020-11-17
add django 3.1 docs

SyedAzeemJaved:
added django 3.1 official documentation to free-programming-books.md

latest version of stable django release

official documentation, absolutely free


eshellman:
Do we want to remove the links for the no-longer-supported versions?

SyedAzeemJaved:
We should remove the links for versions older than 1.8, as they are not even listed on the official website anymore.

PR #5152, opened 2020-11-18
Cryptography section

noobmaster1112:
Crypto101 pdf added from parrot OS website. This book is open source.


eshellman:
It's not helpful to add a category for a single resource. Try to find another category. Or move other cryptography resources to a new section in language agnostic. Also, is this different from the crypto 101 we already have listed?

Issue #5031, opened 2020-11-01
improve vhf/free-programming-books-lint


borgified:
https://github.com/vhf/free-programming-books-lint/blob/master/dist/lint.js#L10

it appears the linter only considers these "languages". we want to expand the list to include other languages/files as well. eg. Vietnamese, C, Java, SQL lists.

looks like https://github.com/vhf/free-programming-books-lint/blob/master/dist/lint.js#L12 determines what language a particular file is, ie. just based on a part of its filename. maybe its possible to expand that list to include vietnamese since that’s not in the first array of langs. (expecting vn)

in our case, [free-courses-vi.md](EbookFoundation/free-courses-vi.md) maybe it should be vn instead to conform to ISO 3166

contributor13:
I would like to fix this issue. Can anyone please suggest what should I do exactly.

eshellman:
@borgified 'vi' is the correct language code (ISO 639-1); ISO 3166 are country codes.
@contributor13 take a look at  https://github.com/vhf/free-programming-books-lint/blob/master/dist/lint.js#L12 and figure out what needs fixing.

kadhirash:
What I came to notice and my understanding is that from the linter code, it excludes the following files: 'README.md', 'CONTRIBUTING.md', 'CODE_OF_CONDUCT.md', 'SUMMARY.md' 

That's why the alphabetization isn't enforced for the above files, but it is for "CODE_OF_CONDUCT-es.md" as seen in #5117.


The langstoAlphabetize var doesn't include 'th' so the Thai resources won't be checked for alphabetization as seen in #3287. 

The other languages added to the repository but not updated to the langsToAlphabetize var won't be alphabetized, such as Thai(th), Farsi(fa_IR), Tamil(ta), etc.

eshellman:
@computamike has contributed a nice fix for pipes to  https://github.com/vhf/free-programming-books-lint/ I've created 3 issues there to address the problems noted in this issue. Folks with Node experience should find these issues pretty easy to fix.

eshellman:
The most recent PR #5207 seems to have missed an alphabetization error - have we messed up the linter by moving the lists into directories?

davorpa:
> @computamike has contributed a nice fix for pipes to https://github.com/vhf/free-programming-books-lint/ I've created 3 issues there to address the problems noted in this issue. Folks with Node experience should find these issues pretty easy to fix.

🎉 Seems all merged/resolved:
[...]

PR #5199, opened 2020-12-03
reorganize files

eshellman:
1. make it easier to find files
2. reduce scrolling to get to README on main view of repo

## What does this PR do?
creates directories for books, courses, casts, more.


eshellman:
looks like this change breaks the linter. seems that #5031 needs a priority bump

eshellman:
seems the linter failure is a bug in the last commit to the linter

borgified:
linter failure is fixed in #5201, so merge that in first, then update this branch to the new master and it should pass

PR #5240, opened 2020-12-29
Remove dead link to "The Standard Python Library"

marios-zindilis:
The link to the book "The Standard Python Library" hosted on effbot.org is dead:

http://effbot.org/librarybook/

I could not find another source that legally hosts that book, so that I could replace it, so I'm removing it.


eshellman:
archive.org has it: https://web.archive.org/web/20200626001242/http://effbot.org/librarybook/

marios-zindilis:
Thanks @eshellman &mdash; I have updated the link.

eshellman:
Thanks!

PR #5247, opened 2021-01-06
Added entry for Solr for newbies

hectorcorrea:
## What does this PR do?
Adds entry for book "Solr for newbies"

## For resources
Book "Solr for newbies"

### Description 
A book on how to get started with Apache Solr.

I thought the content is relevant to programmers. There are a couple of entries already for ElasticSearch but I didn't see anything for Solr.

I wrote it and licensed under the Creative Commons Attribution 4.0 International License.

### For book lists, is it a book? For course lists, is it a course? etc.
It's the materials for a workshop, in book format.

### DIFF
--- a/books/free-programming-books.md
+++ b/books/free-programming-books.md
@@ -215,6 +215,7 @@
 * [Self](#self)
 * [Smalltalk](#smalltalk)
 * [Snap](#snap)
+* [Solr](#solr)
 * [Spark](#spark)
 * [Splunk](#splunk)
 * [SQL (implementation agnostic)](#sql-implementation-agnostic)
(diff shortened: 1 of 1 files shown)


eshellman:
This is a great addition! 
We've been trying to avoid single-item categories. Are there no other free books on Solr? Would this be better under Java? Should we have a category for things like Solr and Elasticsearch? (on the list of things to do is moving MySQL and Postgres into an SQL category.)

hectorcorrea:
I think grouping Solr and ElasticSearch books into a single category ("Search Tools/DBs/Engines"?) would be a good idea. I didn't feel like I had enough authority/knowledge on how the organization was setup to make that kind of change.

eshellman:
Let's create "Search" under "Language Agnostic". We're looking at ways to improve usability of the too-big books page and I think we'll rename "Language Agnostic" to something like "Topics" or "Subjects". Appreciate your help on this.

eshellman:
#5249

PR #5273, opened 2021-01-23
alphabetize added languages

eshellman:
Our linter had not been looking at these languages.

fixing alphabetization errors


eshellman:
courses done....

PR #5679, opened 2021-10-01
Add Mastering Roda to Ruby's section

brunoocasali:
https://fiachetti.gitlab.io/mastering-roda/ Mastering Roda

## What does this PR do?
Add resource Ruby Roda section and a free book (Mastering Roda) to it.

### DIFF
--- a/books/free-programming-books-langs.md
+++ b/books/free-programming-books-langs.md
@@ -174,6 +174,7 @@ That section got so big, we decided to split it into its own file, the [BY SUBJE
 * [Raspberry Pi](#raspberry-pi)
 * [REBOL](#rebol)
 * [Ruby](#ruby)
+  * [Roda](#roda)
   * [RSpec](#rspec)
   * [Ruby on Rails](#ruby-on-rails)
   * [Sinatra](#sinatra)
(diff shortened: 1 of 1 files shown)


eshellman:
It's usually not helpful to add a category for a single resource. Try to find another category, or see if there are other books in our lists that would fit into your category.

books/free-programming-books-langs.md
```diff
@@ -1996,6 +1997,11 @@ That section got so big, we decided to split it into its own file, the [BY SUBJE
 * [Why's (Poignant) Guide to Ruby](http://poignant.guide)
 
 
+#### Roda
+
+* [Mastering Roda](https://fiachetti.gitlab.io/mastering-roda) - Federico Iachetti, Avdi Grimm and Jeremy Evans
```
davorpa, line 2002: Move this into Python section, more or less on line 1974. Lists alphabetization must be preserved.

davorpa (changes requested):
@brunoocasali, I suggest you move into Python section

Sources repo: https://gitlab.com/fiachetti/mastering-roda

books/free-programming-books-langs.md
brunoocasali: Well, why the Python section? Because roda is a ruby gem not a python library https://roda.jeremyevans.net

brunoocasali: Or you just think that is not enough to create a new section?

davorpa: Yes you are right, I fail with the section :weary:, but the line remains the same :wink: .

brunoocasali: I just know this book about roda, so I think that I will resubmit it adding to ruby section :)

eshellman:
Super!

PR #5809, opened 2021-10-03
Add two books in Polish

mdbrnowski:
## What does this PR do?
Add 2 resources.
Fix some indications.

### Description

One book is about Data Structures, other is a set of tasks and tasks editorials from Polish stage of ICPC.

I learned from these two books a lot!

Both of these books are published on official sites.


eshellman:

PR #5886, opened 2021-10-04
Added a new problem set: acmp (RU)

fadyat:
## What does this PR do?
Added acmp 

### Description
Acmp is a Russian competitive programming site where you can solve various algorithmic problems. There are more than 1000 tasks on the site and they are constantly updated. Each of the tasks on some topic can be indicated in the task, for example, binary search or breadth-first search in a graph. There are also sections devoted to the past real Olympiads, which were written by schoolchildren. The problem can be submitted directly on the site, if something went wrong, you will be shown the type of error and the test on which it fell. I think you are familiar with this :)
This site shows interesting and exclusive tasks, both for a beginner in sports programming as well as for a professional

## Checklist:
- [x] Read our [contributing guidelines](https://github.com/EbookFoundation/free-programming-books/blob/master/CONTRIBUTING.md)
- [x] Search for duplicates.
- [x] Add needed indications (PDF, access notes, under construction)

## Follow-up

### DIFF
--- a/more/problem-sets-competitive-programming.md
+++ b/more/problem-sets-competitive-programming.md
@@ -79,6 +79,7 @@
 
 * [500 Data structures and algorithms interview questions and their solutions in C++](https://www.quora.com/q/techiedelight/500-Data-Structures-and-Algorithms-interview-questions-and-their-solutions)
 * [A2 Online Judge](https://a2oj.com/ps)
+* [Acmp](https://acmp.ru)
 * [Advent Of Code](http://adventofcode.com)
 * [Anarchy Golf](http://golf.shinh.org)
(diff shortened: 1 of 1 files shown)


eshellman:
do you think it needs a notaion that it's russian?

fadyat:
> do you think it needs a notaion that it's russian?

in the final file is certainly not needed. there are still sites in some Asian languages, there is no annotation, then we will not break the structure

eshellman:
since the name of the resource gives no hist that it's in russian, I would add a notiation, even if its the first to do so.

fadyat:
is it okey?

eshellman:
The title should be "Школа программиста", I think

PR #6056, opened 2021-10-08
added django course

ichbinmeraj:
added django persian taught course to free-courses-fa_IR


ichbinmeraj:
i fixed things that you mentioned. e.g adding more course to django section and blank lines number between sections.

PR #6062, opened 2021-10-08
added vue.js 3 course

ichbinmeraj:
added vue.js 3 persian taught course to free-courses-fa_IR


eshellman:
It's usually not helpful to add a category for a single resource. Try to find another category, or see if there are other books in our lists that would fit into your category.
For youtube, we prefer "playlist" links and the name of the producer/creator. https://github.com/EbookFoundation/free-programming-books/blob/master/CONTRIBUTING.md#platforms-and-access-notes

eshellman:
that's not the javascript section!

PR #6176, opened 2021-10-11
Slovenian language introduced

BlazOrazem:
- Slovenian language (sl) is introduced in HowTo section
- a filename free-courses-sl.md is renamed to free-courses-si.md (because Sinhala language abbreviation is 'si' and not 'sl' which stands for Slovenian)


eshellman:

PR #6031, opened 2021-10-07
add CONTRIBUTING pt-BR

ricardokojo:
## What does this PR do?
Add CONTRIBUTING-pt_BR.md

### Description

Add pt-BR (Brazilian Portuguese) translation of Contributing guide. Add reference to new Contributing guide in other translation files.

## Follow-up

- Check the status of GitHub Actions and resolve any reported warnings!

Issue #5762, opened 2021-10-02
chore:  hacktoberfest 2021 backlog


davorpa:
When all translations will be stable, we have to remember do some chores after touching common words or translate repo docs.

_Started by @davorpa in https://github.com/EbookFoundation/free-programming-books/pull/5732#discussion_r720701544_

<div align="center" markdown="1">

[![Hacktoberfest 2021](https://img.shields.io/github/hacktoberfest/2021/EbookFoundation/free-programming-books)](https://github.com/EbookFoundation/free-programming-books/pulls?q=is%3Apr+is%3Amerged+created%3A2021-10-01..2021-10-31)

</div>

### Core / Governance

- [x] ci: Get feedback about https://github.com/EbookFoundation/free-programming-books/issues/5176#issuecomment-890391432 / Escape pipes, sharps and other meaningful markdown tokens... in text like headings, resource titles...
- [x] #6590
- [x] #6164. **Improved at**: #6614
- docs: Apply new changes made in english gobernance files (CoC, CONTRIBUTING, HOWTO) to translations
  - [x] Apply #5351  to translations
  - ...
- [x] Consistence format for entry By Lang / By Subject #5602 - #5774
- [x] Title case for product brands #5940, #6384
- [x] #6791
- [x] #6792

### Translation / alphabetization issues

- [x] #6782
- [x] ~Incomplete/partial Indonesian translation #5816 #5898. Maybe #5752 fix it~. Addressed by: #6719
- [x] #6783
- [x] Review #5906 Maybe it uses some Translation service. #6036 
- [x] #7050
- [x] Korean `ko_KR`/`ko_KP` translations #6087. Maybe first commits of #5837 completes some parts of the others (#5908 and #6060 + #6503)
- [x] #5831  https://github.com/EbookFoundation/free-programming-books/commit/d775cb34fbf5eadbdaa465bf46510cc4974151f8#commitcomment-59537594 Minor change in flag emoji markup. **Solved at**: #6619
- [x] Solve review bugs posted after merge https://github.com/EbookFoundation/free-programming-books/pull/6567#pullrequestreview-806520972. **Improved at**: #6618
- [x] #6714 Solve RTL magics linter issues https://github.com/EbookFoundation/free-programming-books/pull/6715#issuecomment-1036637292 https://github.com/vhf/remark-lint-alphabetize-lists/pull/20

### Content / Formating

Improves topic #4579

- [x] Escape Pipes & Typo Trademarks #6577
[...]

PR #6271, opened 2021-10-16
Fix CONTRIBUTING-pt_BR

ricardokojo:
## What does this PR do?
Fixes CONTRIBUTING-pt_BR (comments at #6031)

### Description

Fixes formatting and minor errors pointed at PR #6031.

## Follow-up


ricardokojo:
This should solve the problems pointed at #6031. As I commented in the previous PR, the symbol `¡` - known for its use in exclamations, such as `¡Error!` - is not used in Brazilian Portuguese :brazil: :portugal:.

The main point was solved by #6168.

This should tick one of the items at #5762.

PR #6322, opened 2021-10-19
Add new PHP resources from coderlessons (RU)

ShkiperMe:
## What does this PR do?
Add new PHP resources from coderlessons

PR #6325, opened 2021-10-19
Add new Python resources from coderlessons (RU)

ShkiperMe:
## What does this PR do?
Add new Python resources from coderlessons


davorpa:
More or less like in https://github.com/EbookFoundation/free-programming-books/pull/6322#issuecomment-946828947

PR #6507, opened 2021-10-29
Create translate hindi to english

Wraithh00:
with this it can be converted from one language to another language  that too easily

## What does this PR do?
Add resource(s) | Remove resource(s) | Add info | Improve repo

## Checklist:
- [ ] Read our [contributing guidelines](https://github.com/EbookFoundation/free-programming-books/blob/master/CONTRIBUTING.md)
- [ ] Search for duplicates.
- [ ] Add needed indications (PDF, access notes, under construction)

## Follow-up

### DIFF
--- a/translate hindi to english
+++ b/translate hindi to english
@@ -1,0 +1,6 @@
+from googletrans import Translator
+
+translator = Translator()
+
+out =translator.translator("मेरा नाम गुरप्रीत सिंह है।",dest="en")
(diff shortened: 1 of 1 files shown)


LuigiImVector:
It would be enough to use google translate but in order to maintain a certain level of quality we avoid using it and we rely only on users' translations. Thanks anyway.

Issue #5983, opened 2021-10-06
proposal(docs): repository badges


davorpa:
I've checked for all governance documents and the conclusion is we have the possibility to add:

- Awesome (already present)
- Contributor Covenant (as suggested at [www.contributor-covenant.org](https://www.contributor-covenant.org/#how-to-adopt-contributor-covenant))
- CC license (see [this gist](https://gist.github.com/qvil/5e3ed56c26d784e51424621119cc4028) for other designs)

An example, following other repositories, could be:

![image](https://user-images.githubusercontent.com/3125580/136253819-bf96c416-cc44-45bb-a599-812aff08b3e7.png)

Of course, this is a draft, not pushed into my fork yet. It can be placed at other part or files like CONTRIBUTING or HOWTO

If is taken as is, I'll also opt for remove it from 1st heading.

We reach an agreement among the entire community as to whether it is useful, its place and/or the order of the badges.

LuigiImVector:
I think that's a good idea, maybe the Contributor Covenant badge could be placed in CONTRIBUTING.md

![image](https://user-images.githubusercontent.com/59611597/137194578-8155da58-1a4a-4df3-ad91-0c19807dff3c.png)

davorpa:
More proposed badges at #6697

PR #6631, opened 2021-12-01
Add badges

LuigiImVector:
## What does this PR do?
Improve repo

### Description

Added badges on README and CONTRIBUTING (see #5983)

## Checklist:
- [X] Search for duplicates.

## Follow-up

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,7 +1,5 @@
 *[Read this in other languages](../README.md#translations)*
 
-[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)](CODE_OF_CONDUCT.md)
-
 ## Contributor License Agreement
 By contributing you agree to the [LICENSE](https://github.com/EbookFoundation/free-programming-books/blob/main/LICENSE) of this repository.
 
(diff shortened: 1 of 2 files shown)


davorpa:
I'd preffer all in the same site like in other repos

LuigiImVector:
> I'd preffer all in the same site like in other repos.

@eshellman what do you think?

eshellman:
I'm not a big fan of badges. The introduce a dependency.
For the CC license, if we want a badge we should use the one from Creative Commons. For the "Contributor Covenant", our CoC is "adapted from Contributor Covenant" but we don't use the words "Contributor Covenant" except on the CoC attribution, so I worry that the added language on the README might be confusing.
I would like to move the "Awesome" badge to below the fold. It's nice to have but oddly prominent on our page.

LuigiImVector:
I removed the CoC badge to avoid confusion. The official CC license badge doesn't have the same height of "awesome" badge:

image

I think that the unofficial badge is ok but if it is not so, I kick it out and leave the "awesome" badge under the title.

docs/CONTRIBUTING.md
```diff
@@ -1,6 +1,5 @@
 *[Read this in other languages](../README.md#translations)*
 
-
```
davorpa, line 3: Empty line removed :wink:

PR #6726, opened 2022-02-20
Fix link to Hebrew assembly book

oshaked1:
Previous link led to the python book.

## What does this PR do?
Fix link

Previous link was a duplicate to the python book that is listed as well.

Same domain and author as the OS, python and networking books on the same page.


davorpa (approved):
Nice catch!!

eshellman:

Issue #6107, opened 2021-10-09
What's a programming playground?


eshellman:
in pr #5885 , there's a question about what types of resources should be in the playgrounds list. Comments specific to #5885 should go there, general comments and suggestions for our documentation can go here.

PR #6817, opened 2022-04-05
Added playground definition

Suman373:
Added a small definition about programming/coding playgrounds for making it understandable to beginners

## What does this PR do?
Add Info

It is valuable because it explains what coding playground is in few lines.
## Checklist:
- [ ] Read our [contributing guidelines](https://github.com/EbookFoundation/free-programming-books/blob/main/docs/CONTRIBUTING.md)
- [ ] Search for duplicates.
- [ ] Add needed indications (PDF, access notes, under construction)

## Follow-up

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,8 @@ Free Podcasts and Screencasts:
 
 ### Programming Playgrounds
 
+Whether you're a beginner programmer or an expert developer, code/programming playgrounds are useful when sharing and learning with others. A programming playground is an online service where you can write, compile(or run), and share code with others. They also allows you to fork and get your hands dirty by playing around with other's code.
+
 + [Chinese / 中文](more/free-programming-playgrounds-zh.md)
 + [English](more/free-programming-playgrounds.md)
 


CleanMachine1:
This seems unnecessary to me since a "programming playground"  to me says "mess about with a programming language online with a repl or similar" and also out of place, since the rest of the types of resources do not have a description. 

I think if this were to be implemented into the README, it should be applied to all of the other types of resources, however this would be cluttering and make the README less succinct, which make it seem just like fluff.

Either way, this is way to long, it should be a brief description.

Suman373:
> This seems unnecessary to me since a "programming playground" to me says "mess about with a programming language online with a repl or similar" and also out of place, since the rest of the types of resources do not have a description.
> 
> I think if this were to be implemented into the README, it should be applied to all of the other types of resources, however this would be cluttering and make the README less succinct, which make it seem just like fluff.
> 
> Either way, this is way to long, it should be a brief description.

> Should I shorten the definition instead?

Suman373:
> That would be a good start but I'm still weary with adding this.

> Alright! No problem, I'll close the pr. Lmk if I could fix any other issues !

CleanMachine1:
Lets get the opinion of others before closing, since I am just one of many reviewers :)

Suman373:
> Sure. :))

davorpa:
Related with #6107

eshellman:
I agree this category would benefit from some explanation. The proposed explanation is perhaps too specific - there are some resources that don't quite fit the explanation. In particular, there are some games that would fit well here. The shorter the better. Maybe "try it!" is enough?

Suman373:
> Yeah, I think shortening the explanation and adding "try it " would seem better!

Suman373:
> That could be done as well.

davorpa:
Closed in preference of #6817

Issue #5491, opened 2021-08-18
UI Improvement


Aanchal116:
The current UI seems to be very plain with no headers and footers. I would like to propose working on making the website more attractive.

eshellman:
I assume you mean https://ebookfoundation.github.io/free-programming-books Very little has but done on the design/cosmetics, so there's plenty of room to improve. But can't change our markdown style. PRs welcome, incremental improvements are better than complete overhauls.

Aanchal116:
Yes, you got it right. I'll see how I can maintain the current flow with the improvements. Thankyou.

ShehanMaduwantha:
What if someone makes a nice frontend from these markdown files ?

LuigiImVector:
maybe we can just use another theme, like cayman: https://luigiimvector.github.io/free-programming-books/ (branch [pages-test](https://github.com/LuigiImVector/free-programming-books/blob/pages-test/_config.yml))

PR #6897, opened 2022-07-02
style: change theme

LuigiImVector:
## What does this PR do?
Improve repo

### Description

I changed the theme from minimal to [cayman](https://github.com/pages-themes/cayman), you can see the new page here: https://luigiimvector.github.io/free-programming-books/
The UI wasn't very nice, as said on #5491.

## Checklist:

## Follow-up

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,6 +1,6 @@
 # [Name of visual theme]
-#theme: jekyll-theme-minimal
-remote_theme: pages-themes/minimal@v0.2.0
+#theme: jekyll-theme-cayman
+remote_theme: pages-themes/cayman@v0.2.0
 
 # [Conversion]
 markdown: kramdown


eshellman:
not crazy about the cayman theme, esp on desktop.
I would like to push people towards the dynamic site, which uses minimal theme. 
Need feedback from others!

CleanMachine1:
Just giving my opinion, I prefer the cayman style to the current one

LuigiImVector:
What do you think @davorpa ?

davorpa:
> What do you think @davorpa ?

I prefer the current minimal theme because is the most simple and fast. Maybe change it's styles a bit extending jekyll theme due to official github-themes development are frozen since remote-theme gem was appear

YuvrajSHAD:
for better understanding and experience, i think Cayman theme is good.

Thenlie:
I wonder if we could create a discussion poll or something for this? I like both themes and think it is mostly a toss up. So maybe letting the community decide would be a good option.

eshellman:
yes, should be an issue.
we need verification that the theme will also work on free-programming-books-search

PR #6914, opened 2022-07-13
build(fpb-lint): linting errors as PR comments, resolved #4416

### LINKED ISSUE #4416: Linting Errors as PR Comments
I think it'd be worth pulling the lint summary from CI and posting them as a comment in the PR?

Many contributors either don't know how or don't bother to check the failed checks. And some that do, don't understand what's wrong because they're only reading the bottom line, or just don't know what to focus on.

It'd massively reduce the workload of maintainers to push the warnings to a PR comment, so the relevant information that contributors need is readily available.

LuigiImVector:
## What does this PR do?
Improve repo

### Description

I improved the fpb-lint.yml workflow so that linter errors were shown via comment, as proposed by @SethFalco on #4416.

In this PR LuigiImVector#10 I conducted some tests, all successful.

## Checklist:

## Follow-up

### DIFF
--- a/.github/workflows/comment-pr.yml
+++ b/.github/workflows/comment-pr.yml
@@ -1,0 +1,56 @@
+name: Comment on the pull request
+
+on:
+  workflow_run:
+    workflows: ["free-programming-books-lint"]
+    types:
+      - completed
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  upload:
+    runs-on: ubuntu-latest
+    if: >
+      ${{ github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: 'Download artifact'
+        uses: actions/github-script@v6
+        with:
+          script: |
+            let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: context.payload.workflow_run.id,
+            });
+            let matchArtifact = allArtifacts.data.artifacts.filter((artifact) => {
+              return artifact.name == "pr"
+            })[0];
+            let download = await github.rest.actions.downloadArtifact({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               artifact_id: matchArtifact.id,
+               archive_format: 'zip',
+            });
+            let fs = require('fs');
+            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/pr.zip`, Buffer.from(download.data));
+
+      - name: 'Unzip artifact'
+        run: unzip pr.zip
+      
+      - name: 'Comment on PR'
+        run: |
+          cat error.log
+          if [ -s error.log ]
+          then
+            gh pr review $(<PRurl) -r -b "Linter failed, fix the error(s):
+            \`\`\`
+            $(cat error.log)
+            \`\`\`"
+            gh pr edit $(<PRurl) --add-label "linter error"
+          else
+            gh pr review $(<PRurl) -a
+            gh pr edit $(<PRurl) --remove-label "linter error"
+          fi
--- a/.github/workflows/fpb-lint.yml
+++ b/.github/workflows/fpb-lint.yml
@@ -37,7 +37,7 @@ jobs:
         fpb-lint ./courses/ &>> output.log || echo "Analyzing..."
         fpb-lint ./more/ &>> output.log || echo "Analyzing..."
 
-        cat > error.log
+        touch error.log
 
     - name: Clean output
       if: ${{ always() && 
--- a/.github/workflows/fpb-lint.yml
+++ b/.github/workflows/fpb-lint.yml
@@ -2,11 +2,6 @@ name: free-programming-books-lint
 
 on: [push, pull_request]
 
-env:
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  GITHUB_RUN_ID: ${{ github.run_id }}
-  PR: ${{ github.event.pull_request.html_url }}
-
 jobs:
   build:
 
@@ -87,20 +82,14 @@ jobs:
             });
           });
     
-    - name: Print output
+    - name: Upload artifact
       if: ${{ always() && 
           github.event_name == 'pull_request' }}
       run: |
-          cat error.log
-
-          if [ -s error.log ]
-          then
-            gh pr review $PR -r -b "Linter failed, fix the error(s):
-            \`\`\`
-            $(cat error.log)
-            \`\`\`"
-            gh pr edit $PR --add-label "linter error"
-          else
-            gh pr review $PR -a
-            gh pr edit $PR --remove-label "linter error"
-          fi
+          mkdir -p ./pr
+          echo ${{ github.event.pull_request.html_url }} > ./pr/PRurl
+          mv error.log ./pr/error.log
+    - uses: actions/upload-artifact@v3
+      with:
+        name: pr
+        path: pr/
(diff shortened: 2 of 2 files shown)


.github/workflows/fpb-lint.yml
```diff
@@ -2,19 +2,109 @@ name: free-programming-books-lint
...
   build:
 
     runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v3
+    - name: Use Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
```
davorpa, line 20: Not needed since uses `actions/github-script` to clean output

LuigiImVector: oh right, I forgot to remove it, thanks

.github/workflows/fpb-lint.yml
```diff
@@ -14,7 +19,88 @@ jobs:
...
+    - name: Pull Request
+      if: ${{ always() && 
+          github.event_name == 'pull_request' }}
+      run: |
+        fpb-lint ./books/ &>> output.log || echo "Analyzing..."
+        fpb-lint ./casts/ &>> output.log || echo "Analyzing..."
+        fpb-lint ./courses/ &>> output.log || echo "Analyzing..."
+        fpb-lint ./more/ &>> output.log || echo "Analyzing..."
+
+        cat > error.log
```
davorpa, line 40: Is this line to create a empty file that can be used at line 60?, Better use `touch` I tought

```shell
touch error.log
```

LuigiImVector: You're right, create a code suggestion

.github/workflows/fpb-lint.yml
```diff
@@ -2,6 +2,11 @@ name: free-programming-books-lint
 
 on: [push, pull_request]
 
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GITHUB_RUN_ID: ${{ github.run_id }}
+  PR: ${{ github.event.pull_request.html_url }}
+
 jobs:
   build:
```
davorpa, line 11: About current workflow error at `Print output` task.... 

image

Reading https://www.damirscorner.com/blog/posts/20220506-TokenPermissionsForGithubActions.html... a bit I think that is about GitHub token permissions

See Seth workarround at PR #5557 : https://github.com/SethFalco/free-programming-books/blob/e9f7dff3be87e640a72f97bd4eec1be79ec2b43a/.github/workflows/fpb-lint.yml#L7-L9

SethFalco:
Might be worth evaluating this one more.

The problem I ran into with my one which through me off a bit was that if you look at your tests carefully, you executed all of those comments from a PR.

A PR that could just as easily modify the workflow file to do whatever they want as part of their PR, include spam or execute other actions if given write access to the repo.

Before we merge this, would you mind merging the one in your fork and perhaps I can try do a malicious PR to it and we'll see what happens? 🤔

.github/workflows/fpb-lint.yml
SethFalco: Note, my workaround is insecure and shouldn't be used if I understand correctly.

The real solution is in my comment below. (Split into 2 separate workflows, one creates the message, the other reads the message and posts the review.)

LuigiImVector:
> Before we merge this, would you mind merging the one in your fork, and perhaps I can try to do a malicious PR to it, and we'll see what happens? 🤔

Now I'm going to read the article you linked, in the meantime is it enough for me to write?
`on: [push, pull_request_target]`

I'm not sure if I understand correctly

SethFalco:
I recommend you scroll down to where it says:

> Below is an example of the intended usage in which the results of an unprivileged pull_request workflow are combined with a privileged workflow to leave a comment in response to a received PR:

Over here you'll see an example of it using 2 workflows to be able to leave comments/reviews without creating the security issue.

LuigiImVector:
> Over here you'll see an example of it using 2 workflows to be able to leave comments/reviews without creating the security issue.

_"Incoming data from artifacts is potentially untrusted. When used in a safe manner, like reading PR numbers or reading a code coverage text to comment on the PR, it is safe to use such untrusted data in the privileged workflow context. However if the artifacts were, for example, binaries built from an untrusted PR, it would be a security vulnerability to run them in the privileged workflow_run workflow context. Artifacts resulting from untrusted PR data are themselves untrusted and should be treated as such when handled in privileged contexts."_ 

Even this could not be a totally safe solution

---

Maybe it's enougth to [disable the auto-run](https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-required-approval-for-workflows-from-public-forks) of the workflow, giving us the ability to approve only non-suspicious PR

SethFalco:
> When used in a safe manner, like reading PR numbers or reading a code coverage text to comment on the PR, it is safe to use such untrusted data in the privileged workflow context

> However if the artifacts were, for example, binaries built from an untrusted PR, it would be a security vulnerability to run them in the privileged workflow_run workflow context

You're right, but in this case, it's fine because we're just reading a text file, not executing a binary. Good catch though, it's definitely good to be vigilant just to make sure we're 100% sure we're not screwing this repo up. XD

LuigiImVector:
So I'll try to implement the 2 workflows here

LuigiImVector:
I divided the workflow and now the `free-programming-books-lint` doesn't any return error; `workflow_run` is only triggered in the default branch

.github/workflows/comment-pr.yml
```diff
@@ -0,0 +1,56 @@
...
+  workflow_run:
+    workflows: ["free-programming-books-lint"]
+    types:
+      - completed
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  upload:
```
SethFalco, line 13:
```suggestion
  upload:
    permissions:
      pull-requests: write
```

This job would need to explicitly declare write access for `pull-request` here, I think?
Either that, or in the repo settings someone would have to enable write access by default for all workflows, but that's probably not a good idea for a public repo.

.github/workflows/fpb-lint.yml
```diff
@@ -14,7 +14,82 @@ jobs:
...
+    - name: Pull Request
+      if: ${{ always() && 
+          github.event_name == 'pull_request' }}
+      run: |
+        fpb-lint ./books/ &>> output.log || echo "Analyzing..."
+        fpb-lint ./casts/ &>> output.log || echo "Analyzing..."
+        fpb-lint ./courses/ &>> output.log || echo "Analyzing..."
+        fpb-lint ./more/ &>> output.log || echo "Analyzing..."
+
+        touch error.log
```
SethFalco, line 35:
```suggestion
```

If my other suggestions are merged, these line can be deleted.

SethFalco (commented):
I've tested all the suggestions in this repo:
https://github.com/SethFalco/free-programming-books/pull/6

---

There is another issue. Since we have 2 workflows named `free-programming-books-lint`, one for `push` and one for `pull-request`, the comment action gets triggered twice, with one always failing since the artifacts don't exist.

I don't think we really need this on `push` anyway, so perhaps we can drop that one and make this run on `pull-request` only?

---

Giving it more thought, I originally proposed that we should have the action approve or request. However, I think that might be a bad idea, and we should instead only post a comment.

Since we have many other checks in this repository in other actions, it'd look odd if this action approved the pull request, while other linter actions failed. Contributors will receive contradicting information and won't be sure whether they need to fix something or not. We should probably just post a comment, and later we can try to hide it once it's outdated.

This can be done with the following change:

```diff
-             gh pr review $(<PRurl) -r -b "Linter failed, fix the error(s):
+             gh pr comment $(<PRurl) -b "Linter failed, fix the error(s):
              \`\`\`
              $(cat error.log)
              \`\`\`"
              gh pr edit $(<PRurl) --add-label "linter error"
            else
-             gh pr review $(<PRurl) -a
```

Milad335t (approved):

Milad335t (approved):

PR #6985, opened 2022-08-05
Update free-programming-books-ta.md

kbdharun:
**Fixing ~broken~ http links to all books with links pointing to Wayback machine and Internet archive, press books. Also, added the author's name in Tamil beside each book.**

**Previously the links were pointing to <http://www.kaniyam.com> which is** ~offline and unreachable~ [EDIT] *in http therefore it is blocked by default in most browsers* **currently. So I searched their new website <https://freetamilebooks.com> and over there I found internet archive links to the books, for books not present in the internet archive I used the WayBack Machine to link to the old pdf download page of the book.**

Also, I added the author's name beside each book and specified (PDF).

## What does this PR do?
Improve repo

These books are valuable and would be very useful for students and professionals who can understand concepts in Tamil better than in English. 

I updated the old links to newer ones, seeing most of the books have creative commons licensing I am pretty sure they are free. 

## Checklist:

### DIFF
--- a/books/free-programming-books-ta.md
+++ b/books/free-programming-books-ta.md
@@ -13,45 +13,45 @@
 
 ### BigData
 
-* [எளிய தமிழில் Big Data](https://archive.org/details/learn-bigdata-in-tamil/learn-bigdata-in-tamil) - து. நித்யா
+* [எளிய தமிழில் Big Data](https://archive.org/details/learn-bigdata-in-tamil/learn-bigdata-in-tamil) - து. நித்யா (PDF)
 
 
 ### Database
 
-* [எளிய தமிழில் MySQL ](https://web.archive.org/web/20130120053308/http://www.kaniyam.com/download/Learn-MySQL-in-Tamil-V1.pdf) - து. நித்யா
+* [எளிய தமிழில் MySQL ](https://web.archive.org/web/20130120053308/http://www.kaniyam.com/download/Learn-MySQL-in-Tamil-V1.pdf) - து. நித்யா (PDF)
 
 
 ### HTML and CSS
 
-* [எளிய தமிழில் CSS](https://archive.org/details/learn-css-in-tamil/learn-css-in-tamil) - து. நித்யா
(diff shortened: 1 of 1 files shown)


LuigiImVector:
> Previously the links were pointing to https://www.kaniyam.com which is offline and unreachable currently

kaniyam.com works correctly

kbdharun:
> kaniyam.com

I noticed the links were in HTTP and were blocked by the browser by default. It is possible for me to access it in Firefox. But not in chrome. 
The below image is displayed when I visit that site
image

Wouldn't it be better to link directly to internet archive, then redirecting to this page as it is in HTTP and HTTPS page isn't present.

LuigiImVector:
This problem occurs when using the version with the site's HTTPS (https://www.kaniyam.com/learn-bigdata-in-tamil-ebooks/) while using HTTP, such as the original resource link, does not experience this problem even on Chrome

kbdharun:
> This problem occurs when using the version with the site's HTTPS (https://www.kaniyam.com/learn-bigdata-in-tamil-ebooks/) while using HTTP, such as the original resource link, does not experience this problem even on Chrome

In http, version to access the webpage chrome would ask for your permission twice to access site. In https, version which the browser redirects doesn't have a proper certificate. Which might make people assume that the file isn't secure. 

Since, the original publishers had published it in internet archive wouldn't it be better because the link would be permanent and also secure. And also easy for others to download in other formats like EPUB too.

LuigiImVector:
> Both the cases might make potential readers assume that the file isn't secure.

You are right, there are many HTTP resources in the repository, I think an effective way to make them HTTPS should be discussed, always using archive.org I think can be inconvenient sometimes.

Anyway for me these changes are fine, let's wait for others' opinion.

davorpa:
I'm in pro of leave as http://www.kaniyam.com/learn-bigdata-in-tamil-ebooks/ with `HTTP`

eshellman:
http links are still acceptabe.

PR #6987, opened 2022-08-05
Move an AWS course from books to courses section

AhmedElTabarani:
## What does this PR do?
Fix the mistake of #6752 PR

## Checklist:
- [ ] Read our [contributing guidelines](https://github.com/EbookFoundation/free-programming-books/blob/main/docs/CONTRIBUTING.md).
- [ ] [Search](https://ebookfoundation.github.io/free-programming-books-search/) for duplicates.
- [ ] Add needed indications (PDF, access notes, under construction).
- [ ] Used an informative name for this pull request.

## Follow-up

### DIFF
--- a/books/free-programming-books-ar.md
+++ b/books/free-programming-books-ar.md
@@ -4,7 +4,6 @@
 
 * [Arduino](#arduino)
 * [Artificial Intelligence](#artificial-intelligence)
-* [AWS](#aws)
 * [C#&lrm;](#csharp)
 * [DB & DBMS](#db--dbms)
 * [HTML and CSS](#html-and-css)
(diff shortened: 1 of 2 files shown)


davorpa (approved):

davorpa:
Right addressed PR is #6752

AhmedElTabarani:
@davorpa is there a way to make multiple PR ? if I do 2 PR in difference branch, the second one will hold the previous commit of the first PR (which is in difference branch)

davorpa:
> 

Is discouraged to use incremental PRs sharing same branch (at least here, that we use squash as merge strategy).

Is more an idiomatic and organizative problem than ones of we can do. If you think that two commits should be in the same branch, maybe needs rename PR to grow it scope in the middle of both pushes

AhmedElTabarani:
> Is discouraged to use incremental PRs sharing same branch (at least here, that we use squash as merge strategy).
> 
> Is more an idiomatic and organizative problem than ones of we can do. If you think that two commits should be in the same branch, maybe needs rename PR to grow it scope in the middle of both pushes

I always made a new branch for new changes, but the new branch take the commits of the old branch (different branch names), i think i will figure out the solution

But can you tell me a advice to avoid that the new branch takes the commits of old branch ?

davorpa:
> But can you tell me a advice to avoid that the new branch takes the commits of old branch ?

Easy! Is posible make a branch from any commit. Navigate from your repo timeline using any editor like VSCode or from web

PR #7000, opened 2022-08-13
Register Tamil locale entry (part of #7001)

GH05T-HUNTER5:
Tamil Programming Tutorial


davorpa:
Tamil resources are added in #7001

davorpa (approved):

PR #7017, opened 2022-08-16
Update free-courses-de.md

GH05T-HUNTER5:
HTMl and CSS
Bootstrap
Added

### DIFF
--- a/courses/free-courses-de.md
+++ b/courses/free-courses-de.md
@@ -3,6 +3,8 @@
 * [C](#c)
 * [C++](#cpp)
 * [Haskell](#haskell)
+* [HTML and CSS](#html-and-css)
+    * [Bootstrap](#bootstrap)
 * [Java](#java)
 * [JavaScript](#javascript)
 * [Künstliche Intelligenz](#künstliche-intelligenz)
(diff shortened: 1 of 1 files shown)


courses/free-courses-de.md
```diff
@@ -27,6 +29,17 @@
...
 
+### HTML and CSS
+
+* [CSS lernen](https://youtube.com/playlist?list=PLuBK_vNnGp8ANspdZh_aRAa1InIhFlgm_) - NEW - Vadim
+* [HTML Tutorial Deutsch](https://youtube.com/playlist?list=PLnlqg5o1zhnhVI3t1iTE2oO4QSGpu7EMx) - Markus Reichl
+
+
+#### Bootstrap
+
+* [Bootstrap 4 Tutorial (Deutsch)](https://youtube.com/playlist?list=PLiH_qbTmMNfhcZazOxjK9hFdtRfq_NLld) - Matthias Dangl
```
eshellman, line 40: It's usually not helpful to add a category or subcategory for a single resource. Usually, we try to find another category (such as `HTML and CSS`, or see if there are other books in our lists that would fit into your category. This is a judgement call - I'm going to allow it here for now, but what we don't want to see is a long list of single entry subcategories for every CSS framework on the planet.

GH05T-HUNTER5: > 

Ok

PR #7020, opened 2022-08-17
change upper case to lower cas in fragment ids

eshellman:
## What does this PR do?
the Markdown processor  creates ids for hearding by changing to lower case, changing space to dash, and removing other characters. This PR fixes index links that use capital letters. note- caps are ok in hex-escaped fragments


davorpa (approved):
LGTM. I can't find anymore

PR #7043, opened 2022-08-28
security: GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/check-urls.yml
+++ b/.github/workflows/check-urls.yml
@@ -1,7 +1,12 @@
 name: Check URLs from changed files
 on: [push, pull_request]
+permissions:
+  contents: read
 jobs:
   job:
+    permissions:
+      # Needed for the 'trilom/file-changes-action' action
+      pull-requests: read
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3


.github/workflows/check-urls.yml
```diff
@@ -1,7 +1,12 @@
 name: Check URLs from changed files
 on: [push, pull_request]
+permissions:
+  contents: read
 jobs:
   job:
+    permissions:
+      # Needed for the 'trilom/file-changes-action' action
+      pull-requests: read
```
davorpa, line 9: As first step... I can say that our repository is configured by default with only read permisions for GitHub Actions.

Moreover, can you explain why and no other permissions added at each line?

And finally, `trilom/file-changes-action` is being replaced with `tj-actions/changed-files` because the number of bugs and not mantained since 2020

sashashura: It is great you have the setting set for the repository. I have no way to see it, thus my pull request. It is up to you to decide, but having it explicit in the yml may have some benefits as better portability if you ever decide to reuse it in other repositories and even narrower set of permissions.

The first change sets the global default `contents: read` permission for all jobs (currently there is only one, but I think it is a good idea to have it if at some point of time a new job is introduced). The `pull-requests: read` is added, because the `trilom/file-changes-action` is used in the job. The `actions/upload-artifact` doesn't need write permission for content. Thus, I think no other permissions are needed. Editing workflows is a bit of trial an error process because I'm not aware of a way to test it locally. However I'm committed to fix any permissions if it starts failing after being merged. Just tag me in the PR.

You are already using `- uses: trilom/file-changes-action@v1.2.4`, thus my comment. I think any possible upgrade should be done as in a separate pull request.

davorpa: Is the first time I hear about talk about this permission for trilom action. Googling a bit I found trilom/file-changes-action#129

So, you are right. LGTM :heart:

davorpa (approved):
LGTM

eshellman:
Thanks, and welcome to free-programming-books! Super to have some extra eyes on things like this.

PR #7092, opened 2022-09-18
Remove Curso do React Native implementando CRUD

Jaozim420:
This link is not working


davorpa:
> This link is not working

It doesn't work or not can be visible due to registration? Could you confirm this?. I don't want to sign up to review this deletion

Jaozim420:
> It doesn't work or not can be visible due to registration? Could you confirm this?. I don't want to sign up to review this deletion

doesn't work, I have an account on this site and the link redirects to the main page

davorpa (approved):

PR #7119, opened 2022-09-27
Add new section for Data Structures and Algorithms

hamadhassan:
There is no section for DSA books. This book is not language-specified. It can be used by all developers in any language
\


eshellman:
check the -subjects list

PR #7382, opened 2022-10-03
Added JavaScript, PHP, Go and Swift books

Harshit-RV:
Added books covering PHP, JavaScript, Go and iOS (Swift) from Flavio Copes.

## Follow-up

- Check the latest books from Flavio Copes at https://flaviocopes.com/


eshellman:
we use "HTML" as the format in stead of "online"

davorpa (approved):

PR #7560, opened 2022-10-04
Added three sections under books-hindi

poorvadiwan:
I added three sections -
1.  C
2. HTML
3. Java

in the books section of the repository for the Hindi Language.


eshellman:

PR #8089, opened 2022-10-11
add book for lao language

phuangpheth:
Thanks for reading my pull request ^^.


eshellman:
1. language code for Laotian is `lo` 
2. all of these are blog posts or articles, not books

PR #8196, opened 2022-10-13
added oop course to sinhala free courses

Anuthra1997:


Anuthra1997:
make OOP [java] as a subtopic of java
capitalized the first letters of the name

PR #8365, opened 2022-10-15
Added a new course in Machine learning

nihar2907:
Added a new crash course


LuigiImVector:
this resource is already there

PR #8521, opened 2022-10-16
[Fix] - free-courses-ko.md broken link

Abhiuday03:


eshellman:
Don't invent titles! Use the title given by the resource.

PR #8549, opened 2022-10-17
Java Eclipse

rezawahyu151:
##What can eclipse do?
- Eclipse is a simple application, where we can code in the Java language. why java? Because one of the current trends is still on Android, so the language may still be used.

I create some programs with java eclipse

https://drive.google.com/file/d/1FZZOdxxjfjxU6NBEdLX9jb1rt4uPTmXG/view?usp=sharing


eshellman:
start a fresh branch

PR #8627, opened 2022-10-18
Fixed typo at Victor's email

ArchontisKostis:
## What does this PR do?
Improve repo

There was a typo and Victor's email was written like this: victorfelder at gmail.com

I changed it to victorfelder@gmail.com


ekrof:
I think that was on purpose as to not be picked up by scraper bots

ArchontisKostis:
Oh! Okay, that didn't occur to me! (Oops)
Either way, I'm going to leave PR open just in case, even if what you're saying makes sense.

eshellman:
it was on purpose

PR #8794, opened 2022-10-21
Added Japanese and Urdu language docs.

RiyaBhaskar12:
Added docs in 2 languages.


eshellman:
suggest you separate into 2 PR's one for urdu another for japanese. The urdu will be held until we've accepted an Urdu list. Would also appreciate  review from others with language expertise.

PR #8825, opened 2022-10-22
Add translate for PT_br in "Cloud Computing" and "editors"

all0cer:
## Better translate in words
 Add info
### the word "Cloud" and "Computing" and "editors" is not PT-br,
but "Nuvem" and "Computação" and "Editores de código" is better to understanding


eshellman:
if you change thhe category names, you have to allso chage the index

PR #8858, opened 2022-10-23
add c programing book

avnishkt:
A book on c language programing  
the book name let us C


eshellman:

PR #8861, opened 2022-10-23
Added resource for Competitive programming

Ujala-2110:
Solved issue no #8835


eshellman:

PR #8899, opened 2022-10-24
Carbon

UjjwalAggarwal-1:
follows up #8645


shrn01:
This might be helpful to only the people working on the Carbon compiler since it does not say anything about the language itself?

UjjwalAggarwal-1:
technically yes, but practically you'd learn about the language as well

eshellman:
Seems Carbon is at its earliest stage. normally we don't like single-resource categories, but you have to start somewhere.

PR #8951, opened 2022-10-25
Changed the authors name as in the title of the book

notGman:
The authors name was corrected as int book. 
This helps to easily search and recognize the book.


eshellman:

PR #9005, opened 2022-10-27
Added a small Python course from Cassio Botaro

fcpimenta:
## What does this PR do?
Added a small, but great Python course from Cassio Botaro.

### Description
It is a course designed for beginners in the computing area, the curious, enthusiasts and everyone who wants to understand the entire process of creating an application until its implementation.

This mini-course has already been used by thousands of people during bootcamps in Brazil.

https://github.com/cassiobotaro/do_zero_a_implantacao/blob/main/LICENSE


fcpimenta:
Incorrect order in my commit.

PR #9287, opened 2023-02-22
security: `set-output` cmd deprecated. Use `$GITHUB_OUTPUT` env file

davorpa:
## What does this PR do?
Improve repo

### Description

>To avoid untrusted logged data to use `save-state` and `set-output` workflow commands without the intention of the workflow author we have introduced a new set of environment files to manage state and output.

Starting 1st June 2023 workflows using `save-state` or `set-output` commands via stdout will fail with an error.

More info:
- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

image

It patches the deprecation

## Checklist:

## Follow-up

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -129,17 +129,21 @@ jobs:
         run: |
           echo $INPUT_ISSUES  \
             | jq --compact-output --raw-output 'del(.[] | .[to_entries[] | .key | select(startswith("_"))])'  \
-            | sed -e 's/^/::set-output name=issues::/'
+            | sed -e 's/^/issues=/'  \
+            >> $GITHUB_OUTPUT
           echo $INPUT_ISSUES  \
             | jq --raw-output '. | length'  \
-            | sed -e 's/^/::set-output name=issues-len::/'
+            | sed -e 's/^/issues-len=/'  \
+            >> $GITHUB_OUTPUT
 
           echo $INPUT_PRS  \
             | jq --compact-output --raw-output 'del(.[] | .[to_entries[] | .key | select(startswith("_"))])'  \
-            | sed -e 's/^/::set-output name=prs::/'
+            | sed -e 's/^/prs=/'  \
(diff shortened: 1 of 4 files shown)


davorpa:
Doing some tests

davorpa:
> Doing some tests to know how to fix awesomebot GitHub summary report
> 
> Compare:
> 
> * OK: [EbookFoundation/free-programming-books/actions/runs/4228037900](https://github.com/EbookFoundation/free-programming-books/actions/runs/4228037900)
> * Bugged: [EbookFoundation/free-programming-books/actions/runs/4242565403](https://github.com/EbookFoundation/free-programming-books/actions/runs/4242565403)

It works again: https://github.com/EbookFoundation/free-programming-books/actions/runs/4243200736

.github/workflows/stale.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           stale-pr-label: " "
 
       - name: Print outputs for issues
-        run: echo ${{ join(steps.stale-issues.outputs.*, ',') }}
```
Thenlie, line 84: Is this change related to the `set-output` deprecation?

Thenlie: I see now this is from https://github.com/actions/stale/pull/859

Thenlie (approved):
This all looks good to me. ✅ 
- No instances of `set-output` or `save-state` were missed in the workflow files. Verified with spotlight search of the repo. 
- Workflow checks are passing and output is properly formatted.
My only nit would be to squash these test commits into one.

.github/workflows/stale.yml
davorpa: Not is related, but since it is a superfluous change, I have taken advantage of this PR to do it.

It should be do it past in #9242 bump

PR #9466, opened 2023-07-19
Add JavaScript with Classes EPUB

diogoeichert:
Add JavaScript with Classes

## What does this PR do?
Add resource(s)

## For resources
JavaScript with Classes

### Description
A simple, yet effective classical OOP approach to JavaScript. From the preamble: “This book is dedicated to programmers who want to adopt the classical OOP paradigm in JavaScript, whether using or not external frameworks and other libraries, as this knowledge pertains to the language level. These techniques take some discipline to master at first, but they prompt the programmer to do more with less effort later on. They may also make the programmer feel more capable of tackling problems for which JavaScript would not even be considered as an option, in a first glance.”

The book approaches JavaScript from a classical OOP perspective, which might be a good (if not the only) way to design some software solutions.

I'm the author, and the book has always been free and it will keep like that for the foreseeable future.

### For book lists, is it a book? For course lists, is it a course? etc.
It's book, it's not a course.

## Checklist:

## Follow-up


eshellman:
Thanks for making you book free! If it has an Open License, consider adding it to Unglue.it and Archive.org.

PR #9534, opened 2023-09-26
Added Jadi Lpic-1 Course

Ar-dst:
I just Add Jadi Lpic-1 Course


eshellman:
Don't invent titles! Use the title given by the resource. It appears that the title for this course is "دوره الپیک ۱ - جادی | LPIC-1 with Jadi". If you use that, escape the |

PR #9607, opened 2023-10-01
Update problem-sets-competitive-programming.md

IRFANSARI:


eshellman:
Not sure I like a site that requires a mobile phone number to try it out. Sorry, I think we'll pass. Maybe we should have a more explicit policy. Maybe should discuss as an issue.

PR #9843, opened 2023-10-04
Removed all broken links for the Free Programming Books AR

3R1Dev:
## What does this PR do?
This PR removes broken links in the Free programming books AR (Arabic).

### Description
The links in the free programming books AR list were all opened to make sure they work and lead to the book they are referencing. Any broken links were removed and the index updated as well, on some the index was removed as there were no books listed.

This was done as requested by issue #9452 so that the issue can be closed.

This is valuable in order to have up-to-date links and correct information.

## Checklist:
Only the links were removed, the quality of the existing references was not tested.


eshellman:
thanks for checking these

PR #9910, opened 2023-10-05
 Added C books

radhesh1:
Added Two C books by Dennis Ritchie and K. Joseph Wesley, R. Rajesh Jeba Anbiah on langs page


eshellman:

Issue #9908, opened 2023-10-05
[Bug]: Search button works without a search text in the search box. Needs 'required' field.


Sark-07:
**Describe the bug**
In the home page, the search bar is working even with blank texts i.e., "." It is taken to another page and an error of "Couldn't get data. Please try again later" is showing.

**To Reproduce**
1. Go to: https://ebookfoundation.github.io/free-programming-books/
2. Click on search button without filling in the search box

**Expected behavior**
If the search box is empty then by clicking on the search button it should say "This field is required" and should not take the user to another page.

**Screen Shot**
![image](https://github.com/EbookFoundation/free-programming-books/assets/85283070/9fa7e595-b970-478d-ab81-b00e5c1c4ec3)

**Video**

https://github.com/EbookFoundation/free-programming-books/assets/85283070/1fb780f3-88e6-440f-8eff-5e737dfb34a1

PR #9924, opened 2023-10-06
Added the required attribute in the search box.md

debanjan-G:
Now the search box has a 'required' field which means that it must have a non-blank value to be valid.

## What does this PR do?
This PR has added the 'required' field in the search input field which makes it mandatory for it to have a non-empty value to be valid. Basically, making the search field required makes it mandatory for the user to enter a value.
It fixes issue #9908.

##After merging this PR, Expected behavior:
If the search box is empty then by clicking on the search button it should say "This field is required" and should not take the user to another page.


eshellman:
duplicate

PR #10068, opened 2023-10-09
Update README.md

GAUTAMMANAK1:
I've removed the extra square brackets that were around the website URL, which were causing the issue. With this corrected code, the badge should work as expected, and when clicked, it will take you to the website provided in the URL.

It seems like you want to create a clickable badge that redirects to a website when clicked. However, there's a small issue with the provided Markdown code. You can correct it by using the following Markdown code:


eshellman:
please describe how to repproduce the bug

GAUTAMMANAK1:
when we scollr the link then all text hover i will change the this bug 
current only blue line hover

3R1Dev:
There are some issues in the pushed changes, listed as follows:

1. On line 13 the text for the ebookfoundation.github.io is being removed and replaced with "Website"
2. line 25 has an unnecessary space added after the URL
3. Line 144 was completely removed

I don't see any issues with the current markdown as it is on the present README. Links work, hover over links works without issue. I dont see the reason for the changes proposed or any bugs.

eshellman:
couldn't reproduce the issue

eshellman:
thanks @3R1Dev for help

PR #10566, opened 2023-10-20
Updated courses-hi.md

404Nikhil:
added resources


kn1-gh7:
> added resources

I would suggest removing the tracking parameter
(The '&si=Oupzc1WR6zbaEytD' part of the link)

PR #10572, opened 2023-10-20

letohx:
Formatted the section headings in the Markdown format for better readability.


casts/free-podcasts-screencasts-ru.md
```diff
@@ -10,11 +10,11 @@
 * [Java](#java)
     * [Spring](#spring)
 * [JavaScript](#javascript)
+    * [React.js](#reactjs)
+    * [Node.js](#nodejs)
```
eshellman, line 14: Node is alphabetically before React.

eshellman (changes requested):
the sections should be reordered to reflect the reordered index

PR #11073, opened 2024-02-27
fix: remove invalid symbols and keep them consistent with the original

ruishawn:
Update docs/CONTRIBUTING-zh.md, remove invalid symbols and keep them consistent with the original.


eshellman:

PR #11084, opened 2024-03-05
Add Chinese courses

ruishawn:


courses/free-courses-zh.md
```diff
@@ -24,4 +24,5 @@
 ### OS
 
 * [操作系统原理](https://www.youtube.com/playlist?list=PLkl2qqmYigA66rJ4FgmZan4YIVRgNFLQx) - 从0开始数
+* [操作系统原理 北京大学](https://www.coursera.org/learn/os-pku) - Professor Chen Xiangqun
```
eshellman, line 27: Note: We do not include titles or honorifics such as "Professor" in our lists.
This course appears to be time-limited. We do not include such courses in our lists, because they would need to be removed after they are done.

ruishawn: Thank you @eshellman, Now i know this note.

PR #11086, opened 2024-03-05
Add Chinese courses about Python

ruishawn:


courses/free-courses-zh.md
```diff
@@ -4,6 +4,7 @@
 * [Flutter](#flutter)
 * [Linux](#linux)
 * [OS](#OS)
```
eshellman, line 6: missed this before; markdown generated ids are lower-case.
```suggestion
* [OS](#os)
```

courses/free-courses-zh.md
```diff
@@ -4,6 +4,7 @@
...
+* [Python](#Python)
```
eshellman, line 7:
```suggestion
* [Python](#python)
```

eshellman (changes requested):

PR #11134, opened 2024-05-02
Update name of the translator

quangnh89:
I myself translated 'The Little Go Book' into Vietnamese. I am only updating it to correct the translator's name.

### DIFF
--- a/books/free-programming-books-vi.md
+++ b/books/free-programming-books-vi.md
@@ -6,7 +6,7 @@
 
 ### <a id="golang"></a>Go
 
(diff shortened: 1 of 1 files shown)


HSUYATANG:
So is it to change the Vietnamese name of this program? Because it looks the same as English.

So how do you think it's different from the previous watch?

quangnh89:
> So is it to change the Vietnamese name of this program? Because it looks the same as English.
> 
> So how do you think it's different from the previous watch?

I want to keep the original name of this book. I just translated content.

eshellman:
Please accept the suggest change to your PR if you agree with it.

quangnh89:
LGFM. What should I do now?

eshellman:
you should have a button to accept the suggestion

PR #11229, opened 2024-08-24
[NO REVIEW] Testing PR #11224

Thenlie:
## What does this PR do?
Improve repo

### Description

Trying to debug why the download artifact step is failing on #11224. 

After seeing this issue https://github.com/actions/download-artifact/issues/339 it was suggested to remove the `name` property from the download step. When I did this, the markdown files would successfully download. See the run here https://github.com/EbookFoundation/free-programming-books/actions/runs/10540794299/job/29205861615?pr=11229. The yml file still failed to download, but this might be fine since it is not a file that will be regularly updated. I _think_ this will unblock the PR in question, but we may want to do a bit more digging to ensure it works for all files. 

## Checklist:
- [ ] Read our [contributing guidelines](https://github.com/EbookFoundation/free-programming-books/blob/main/docs/CONTRIBUTING.md).
- [ ] [Search](https://ebookfoundation.github.io/free-programming-books-search/) for duplicates.
- [ ] Add needed indications (PDF, access notes, under construction).
- [ ] Used an informative name for this pull request.

## Follow-up


eshellman:
i really appreciate your help on this!

.github/workflows/check-urls.yml
```diff
@@ -104,7 +104,7 @@ jobs:
       - name: Checkout  # for having the sources of the local action
         uses: actions/checkout@v4
         # download and unzip the ab-results-*.json generated by job-matrix: check-urls
-      - uses: actions/download-artifact@v4
+      - uses: actions/download-artifact@v3
```
eshellman, line 107: so this fixes it?

Thenlie: It looks like it. I tried some other versions suggested in the issue like `v4.1.8` and those did not work. I'm still a little unclear on the underlying issue, but I think this could be a fine workaround while we keep an eye on the issue for an upstream fix. I will open a new PR with this change only.

Thenlie: https://github.com/EbookFoundation/free-programming-books/pull/11230

PR #11264, opened 2024-09-29
Update check-urls.yml

iceybubble:
Enhanced the code for better performance optimization and security


eshellman:
Code is reformatted to a less readable style

Issue #11260, opened 2024-09-26
Proposed new feature : License notations


eshellman:
One reason the Free Ebook Foundation maintains this repo is to support creators who want to make their work free to others. We especially want to support the use of free licenses, such as those from Creative Commons. One way to do this is allow license notations on free-licensed items.

The easiest way to do this would be to add the license after the end of the item. For example, instead of:

`* [Structure and Interpretation of Computer Programs](https://web.mit.edu/6.001/6.037/sicp.pdf) - Harold Abelson, Gerald Jay Sussman, Julie Sussman (PDF)`

we would have

`* [Structure and Interpretation of Computer Programs](https://web.mit.edu/6.001/6.037/sicp.pdf) - Harold Abelson, Gerald Jay Sussman, Julie Sussman (PDF). CC BY-SA.`

or maybe

`* [Structure and Interpretation of Computer Programs](https://web.mit.edu/6.001/6.037/sicp.pdf) - Harold Abelson, Gerald Jay Sussman, Julie Sussman (PDF; CC BY-SA)`

Discussion welcome, but we should decide by Sept. 30.

PritamP20:
Where should we change these, can you explain if possible

eshellman:
Looks like this will be
`* [Structure and Interpretation of Computer Programs](https://web.mit.edu/6.001/6.037/sicp.pdf) - Harold Abelson, Gerald Jay Sussman, Julie Sussman (PDF) (CC BY-SA)`
to satisfy the linter

eshellman:
@PritamP20 
> Where should we change these, can you explain if possible

testing this at #11269. Comments welcome.

eshellman:
> It might be sensible to use [SPDX identifiers](https://spdx.org/licenses/). For example [`CC-BY-4.0`](https://spdx.org/licenses/CC-BY-4.0.html) (as opposed to `CC BY-4.0`

@edent From the user point of view, the precise version is not really helpful in a listing like ours - the actual version should be specidied in the source, and is frequently unspecified, because the differences in the versions are edge cases. There are also langurage versions if you want to get precise. 

Using SPDX identifiers invites the use of licenses that are unfamiliar to users or not free and the reason I'd like to see us include licenses is to promote the use of licenses that can be relied on without consulting lawyers!

Creative Commons does not use a dash after "CC", I think we should respect that choice.

eshellman:
[...]

PR #11269, opened 2024-09-30
testing the license label

eshellman:
tests #11260

## What does this PR do?
Adds a license label. Does it pass tests?


Thenlie:
Is any additional linting required here? Do we want a `check-license` action similar to `check-urls`?

eshellman:
I don't think a check-license script is practical. In any case, when this is launched, we'll learn more

PR #11635, opened 2024-10-21
Added The Art of the Metaobject Protocol Book

martian2001:


eshellman:
This is only the introduction to the book

PR #11662, opened 2024-10-23
Updated free-courses-en.md with a Github repository 

LionessLilo:
It covers projects from 42 schools. Many of the resources are created by students and are a great learning resource.


eshellman:
In its current form, it's not a course. With a better explanatory page and description, it might work in our problem sets list.

PR #11685, opened 2024-10-26
To add a dark mode toggle button

piyahub:
What does this PR do?
Addition of a dark mode toggle button would be beneficial as it will improve the user experience.
I have made a sample of how it will work and how the toggle button will look like.
Please accept my request.


eshellman:
look at free-programming-books-search; there was an effort to add dark mode but I don't remember what happened.
can't test if the build is failing

PR #12036, opened 2025-10-01
Fix linter summary output

Vidhan0206:
The linter summary now prints to error output (stderr) instead of normal output.


eshellman:
resubmit whe Hacktoberfest is over

PR #12246, opened 2025-10-05
Regex fix: Improve INLINE_CODE_RE to avoid false RTL/LTR warnings

Amitverma0509:
## What does this PR do?
Improve repo (fix regex handling for inline code detection)

## IMPORTANT
- [x] Is this a revision of a previously submitted PR? If so, STOP! Go back, reopen the PR, and add commit(s) the branch you previously submitted. Please don't make the job of reviewing more difficult by hiding previous work.

### Description
Fixes the INLINE_CODE_RE regex in rtl_linter.py so that inline code within text (e.g., Use the `npm install` command.) is properly detected and ignored during RTL/LTR linting.

Prevents false RTL/LTR warnings caused by inline code segments not being recognized, improving the accuracy of lint checks.

N/A – this is a code improvement, not a content addition.

## Checklist:

## Follow-up
Verified that GitHub Actions run successfully after the change

### DIFF
--- a/scripts/rtl_ltr_linter.py
+++ b/scripts/rtl_ltr_linter.py
@@ -105,7 +105,7 @@ HTML_DIR_ATTR_RE = re.compile(r"dir\s*=\s*(['\"])(rtl|ltr)\1", re.IGNORECASE)
 SPAN_DIR_RE = re.compile(r'<span[^>]*dir=["\'](rtl|ltr)["\'][^>]*>', re.IGNORECASE)
 
 # Regex to identify inline code (text enclosed in single backticks)
-INLINE_CODE_RE = re.compile(r'^`.*`$')
+INLINE_CODE_RE = re.compile(r'`[^`]+`')
 
 # Regex to identify the start of a code block (```)
 # Can be preceded by spaces or a '>' character (for blockquotes)


eshellman:
@gabriele-ciccotelli could you review?

gabriele-ciccotelli:
Sure, as soon as I can, after work, I will do some tests and verify the change

gabriele-ciccotelli:
Hi, first of all, I would like to thank @Amitverma0509 for the suggestion.

The new regex proposed is more flexible and correctly identifies inline code snippets even when they are mixed with other text.
However, for the specific purpose of this linter, the original regex was intentionally designed to be stricter and match only segments that consist **entirely** of inline code.

The regular expression, in fact, is used in an `if` check that skips and ignores segments that match, and we want to skip only segments that are purely inline code. For segments that contain a mix of normal text and inline code, like the Arabic equivalent of the phrase “`printf()` the print function” (which is the second one in this image):

image

we actually want the linter to analyze them, we don't want to skip them. These mixed-text segments are candidates for BIDI display issues. If we were to skip them, the linter would miss rendering errors like this:

image

The linter is designed to catch exactly this kind of problem, so that a developer can fix it by adding the appropriate directionality marker (in this case, `&lrm;`):

image

This correction ensures the text is displayed correctly:

image

So, although the new regular expression is valid for finding all occurrences of inline code, in this specific context the idea was to let cases of mixed text pass through in order to analyse them.

Amitverma0509:
So this will be merge in the code?I Guess 😁

gabriele-ciccotelli:
Perhaps I have explained myself wrongly, but with the proposed new regex, segments with inline code mixed with normal text would also be matched and ignored (e.g., the second bullet point in the example I gave in the previous comment) and this must not happen.
The original regex is specifically meant to skip only those segments where there is exclusively inline code and not plain text as well.
So, the current regex is very strict for a specific reason and should remain as it is in this case.
Using the proposed new regex, in the example I showed in the previous comment, the second segment would be matched, skipped and would not be flagged as an error.

eshellman:
thank you both for this effort!

PR #12429, opened 2025-10-12
Fix: scripts- Incorrect <div> Tag Handling logic

Bhavishya2912:
Bug Fix : This PR fixes a bug in `rtl_ltr_linter.py` where the script incorrectly processes `<div> `tags.

**The Bug:**
The lint_file function uses continue statements after detecting an opening or closing `<div>` tag. This causes the linter to stop processing the rest of the line, which leads to a problem if a single line contains both an opening and a closing tag, like `<div dir="rtl">Some Text</div>.`

This bug can make the `block_context_stack` inconsistent, which may result in incorrect linting of subsequent lines and lead to false positives.

**The Fix:**
I have refactored the code to handle all `<div>` tags on a single line. Instead of using continue statements, the script now finds all `<div>` tags on a line and processes them in order, ensuring the `block_context_stack` remains accurate. This fix is a minor change that improves the linter's reliability without affecting its core functionality.

### DIFF
--- a/scripts/rtl_ltr_linter.py
+++ b/scripts/rtl_ltr_linter.py
@@ -247,16 +247,23 @@ def lint_file(path, cfg):
         # Skip lines that start a code block (```)
         if CODE_FENCE_START.match(line): continue
 
-        # Corrected logic to handle all div tags on a single line
+        # Find all opening and closing <div> tags on the line to handle cases
+        # where there can be multiple <div> opening and closing on the same line
         div_tags = re.findall(r"(<div[^>]*dir=['\"](rtl|ltr)['\"][^>]*>|</div>)", line, re.IGNORECASE)
+
+        # Process each found tag in order to correctly update the context stack
         for tag_tuple in div_tags:
+            # re.findall with multiple capture groups returns a list of tuples:
+            # tag: The full matched tag (e.g., '<div...>' or '</div>')
+            # direction: The captured direction ('rtl' or 'ltr'), or empty for a closing tag
             tag, direction = tag_tuple
+
+            # If it's an opening tag with 'markdown="1"', push the new context
             if tag.startswith('<div') and 'markdown="1"' in tag:
-                new_div_ctx = direction.lower()
-                block_context_stack.append(new_div_ctx)
+                block_context_stack.append(direction.lower())
+            # If it's a closing tag and we are inside a div, pop the context
             elif tag == '</div>' and len(block_context_stack) > 1:
                 block_context_stack.pop()
-
         # Check if the line is a Markdown list item
         list_item = LIST_ITEM_RE.match(line)
 


MohitMehtre:
Closes #12430

eshellman:
@gabriele-ciccotelli ?

Bhavishya2912:
Hi @MohitMehtre , it shows all checks have passed. But not merged yet. Anything to do from my side?

gabriele-ciccotelli:
Thank you @eshellman for the ping. I apologize for the delay in responding.

Hi @Bhavishya2912 and thank you for your contribution. I confirm that this would make the linter more accurate and reliable.
Adding a check for unclosed `<div>` tags at the end of the file is also an improvement. 

Just to be sure, have you had a chance to test this new logic on some edge cases? For example:

* A file with incorrect `<div>` structures (such as an unclosed tag or a line with both opening and closing tags) to confirm that the linter correctly identifies all errors?
* Some of the existing, correctly formatted files to ensure that this change does not introduce false positives?

Bhavishya2912:
Hi @gabriele-ciccotelli 
Yes I tested the enhancement with 3-4 files in my local. Raised this issue and PR only after testing the enhanced logic.

scripts/rtl_ltr_linter.py
```diff
@@ -247,21 +247,16 @@ def lint_file(path, cfg):
...
+        # Corrected logic to handle all div tags on a single line
+        div_tags = re.findall(r"(<div[^>]*dir=['\"](rtl|ltr)['\"][^>]*>|</div>)", line, re.IGNORECASE)
+        for tag_tuple in div_tags:
+            tag, direction = tag_tuple
+            if tag.startswith('<div') and 'markdown="1"' in tag:
+                new_div_ctx = direction.lower()
+                block_context_stack.append(new_div_ctx)
+            elif tag == '</div>' and len(block_context_stack) > 1:
+                block_context_stack.pop()
+
```
gabriele-ciccotelli: If I may, I'd also suggest adding the following comments to the new logic:

```suggestion
        # Find all opening and closing <div> tags on the line to handle cases
        # where there can be multiple <div> opening and closing on the same line
        div_tags = re.findall(r"(<div[^>]*dir=['\"](rtl|ltr)['\"][^>]*>|</div>)", line, re.IGNORECASE)

        # Process each found tag in order to correctly update the context stack
        for tag_tuple in div_tags:
            # re.findall with multiple capture groups returns a list of tuples:
            # tag: The full matched tag (e.g., '<div...>' or '</div>')
            # direction: The captured direction ('rtl' or 'ltr'), or empty for a closing tag
            tag, direction = tag_tuple

            # If it's an opening tag with 'markdown="1"', push the new context
            if tag.startswith('<div') and 'markdown="1"' in tag:
                block_context_stack.append(direction.lower())
            # If it's a closing tag and we are inside a div, pop the context
            elif tag == '</div>' and len(block_context_stack) > 1:
                block_context_stack.pop()
```

This will help ensure the code remains easy to understand and maintain in the future.

Bhavishya2912:
Done, updated the file

eshellman:
thank you @gabriele-ciccotelli @Bhavishya2912 for collaboration!

PR #12646, opened 2025-10-19
Add books about testing, both in English and French

Zenigata:
## What does this PR do?
Add free books

https://fr.agilitest.com/resources/white-papers

## Follow-up


eshellman:
these look like file sharing links

PR #12747, opened 2025-10-21
feature/pt br hacktoberfest: Add Python and SQL tutorial links

surlogu:
Add Python and SQL tutorial links


eshellman:
these go in the interactive tutorial list

surlogu:
Python and SQL tutorial links moved to interactive tutorial list

PR #12852, opened 2025-10-26
Add "Pensare in Python" to Italian Python books

Alessio-Falcini:
Added the Italian translation of "Think Python" by Allen B. Downey to the Italian Python books section.


eshellman:

PR #12866, opened 2025-10-27
Enhance README readability using collapsible <details> sections

vasu-devs:
Improved the readability of the README by adding <details> and <summary> tags to collapse long sections. This change helps users navigate the document more easily and access specific sections without scrolling through the entire content.

Why is this valuable?

The <details> structure makes the README cleaner and more user-friendly, especially for large lists. It preserves all content while improving visual organization and accessibility.

How do we know it’s really free?

All existing links and resources remain unchanged — only presentation was updated.


eshellman:
we prefer the native markdown

PR #12978, opened 2025-11-01
Fix: Handle parentheses in author names in RTL/LTR linter

DealsBeam:
The `BOOK_ITEM_RE` regex in `scripts/rtl_ltr_linter.py` did not correctly parse author names containing parentheses. This change updates the regex to correctly capture the author and metadata, even when the author's name includes parentheses.


eshellman:
do you have an example of an author with parentheses?

Issue #12901, opened 2025-10-28
Broken/Malformed link in APL section - College Mathematics with APL


mekapilgupta:
**Link:** https://www.softwarepreservation.org/projects/apl/Books/CollegeMathematicswithAPL
**Section:** APL in free-programming-books-langs.md
**Issue:** Link displays 404 error-> The requested content does not exist.
**Book:** Introduction to College Mathematics with A Programming Language by E.J. LeCuyer
Link Location : https://ebookfoundation.github.io/free-programming-books/books/free-programming-books-langs.html

Cintu07:
hey can i work on this? please assign this issue to me

mekapilgupta:
Can i be assigned this issue as i had reported it.

Xuaya:
Hi, I’d like to work on this issue. Could you please assign it to me? Thanks!

PR #13063, opened 2026-01-12
Fix broken APL book link

Kirti1234-hash:
Added working link for "Introduction to College Mathematics with APL" as reported in #12901


eshellman:
link does not work

PR #13129, opened 2026-03-17
Fix broken link: ASP.NET MVC Music Store (CodePlex shutdown)

24f3001971:
CodePlex was shut down in 2021, making this link permanently dead.

- Old: http://mvcmusicstore.codeplex.com
- New: https://learn.microsoft.com/en-us/aspnet/mvc/overview/older-versions/mvc-music-store/mvc-music-store-part-1

The Microsoft Learn page hosts the same official tutorial content previously on CodePlex.


eshellman:

PR #13143, opened 2026-03-21
fix: remove broken link for Grails Tutorial for Beginners

23f3003643:
The link for "Grails Tutorial for Beginners" (grails.asia) 
is returning a 404 / not loading. Removing the broken link.


eshellman:
link is loading now

PR #13227, opened 2026-04-20
Add Python Machine Learning book

ManjunathByadagi:
Added a Python machine learning book under Python section.


ManjunathByadagi:
I have addressed the linter error by fixing the alphabetical ordering in the Python section. All checks are now passing. Thank you for your review.

eshellman:
These are just code examples from a book, not the book. We don't list those.

PR #13380, opened 2026-07-24
Add free Spanish programming courses (Certmundo)

guilhermecarlos:
Adding four free Spanish-language programming courses from Certmundo (Python, JavaScript, SQL, Git) — each free to study with an optional certificate, same format as the existing Edutin/Platzi entries.


eshellman:

PR #13387, opened 2026-08-01
ci: add names to changed-files output steps for readability

MGPOCKY:
## Problem

In `.github/workflows/check-urls.yml`, the `set-files` step had no `name:`, so a long `run` block shows up as a generic label in the Actions UI.

## Changes

Semantics are unchanged: same `id: set-files`, same `GITHUB_OUTPUT` keys (`files`, `files-len`), and the same job-level output wiring.

- Add `name: Set changed-files outputs` to the existing `set-files` step
- Add `name: Build check-urls matrix` to the existing `set-matrix` step

## Test plan
- [ ] Confirm job outputs still resolve `steps.set-files.outputs.files` and `steps.set-files.outputs.files-len`
- [ ] Open a PR that touches a `.md` / `.yml` file and confirm `check-urls` still runs

### DIFF
--- a/.github/workflows/check-urls.yml
+++ b/.github/workflows/check-urls.yml
@@ -26,7 +26,7 @@ jobs:
     outputs:
       fetch-depth: ${{ steps.set-params.outputs.fetch-depth }}
       files:       ${{ steps.set-files.outputs.files }}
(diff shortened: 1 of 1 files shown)


daltino (commented):
Thanks for working on this! The changes look neat and well-structured.

PR #13421, opened 2026-08-23
fix(id): remove dead Niagahoster Bootstrap 5 tutorial link

mamad2411:
The Niagahoster Bootstrap 5 tutorial link in `books/free-programming-books-id.md` no longer works.

It redirects twice (`niagahoster.co.id` → `hostinger.co.id` → `hostinger.com/id`) and the final destination returns HTTP 404. There's also no archived snapshot available on the Wayback Machine, so per the CONTRIBUTING guidelines I removed the dead entry instead of trying to replace it.


eshellman:

